PR #2, opened 2019-11-08
Add fallback env values

wass3r:


wass3r:
Will revisit in new PR

PR #5, opened 2019-11-09
chore: add caching for actions

wass3r:
Should cache the node_modules folder and make for quicker build runs. Changed `ci` to `i` to prevent `node_modules` getting wiped.


wass3r:
will revisit

PR #11, opened 2019-11-11
Bugfix, Add org button now adds only the filtered repos [ready]

davidvader:


davidvader:
@wass3r ready for review

cypress/fixtures/source_repos.json
```diff
@@ -1,5 +1,81 @@
 {
     "vela": [
+        {
+            "id": 23,
+            "user_id": 9,
+            "org": "vela",
+            "name": "api-2",
+            "full_name": "vela/api-2",
+            "link": "https://github.com/go-vela/api-2",
```
kneal: Not a big deal but in all of the backend components we use `github/octocat` as the default test data.

davidvader: resolved, changed `vela` org to be `github` and made repos `octocat-...`

PR #17, opened 2019-11-13
feat(settings): add repo settings page

davidvader:


davidvader:
See here for checkbox and radio button accessibility information and how tabbing between radios/checkboxes works

https://stackoverflow.com/questions/14322564/can-you-tab-through-all-radio-buttons/14322755

davidvader:
not adding aria labels to checkboxes
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

PR #20, opened 2019-11-14
Feature: rendering server/agent errors for build and steps

davidvader:


src/elm/Build.elm
```diff
@@ -114,25 +114,26 @@ viewBuildItem now org repo build =
             animationStatusClass build.status
 
         markdown =
```
davidvader: here i am cleaning up mis-use of popsicle style. live and learn!

src/elm/Build.elm
```diff
@@ -523,15 +570,15 @@ toTwoDigits int =
 stepClasses : Step -> Steps -> Html.Attribute msg
 stepClasses step steps =
     let
```
davidvader: this fix is needed because I added sorting in steps (server returns them reversed)

src/scss/style.scss
```diff
@@ -1156,8 +1166,8 @@ button.-solid {
 }
 
 .step .summary {
```
davidvader: tighten up the steps, looks cleaner

PR #38, opened 2019-12-05
perf(refresh): better logic when auto-refreshing hooks 

davidvader:


davidvader:
this adds an enhancement to the hooks page so that only hooks being “Viewed” are refreshed by app

PR #80, opened 2020-01-31
feat(logs): link to ranges of log lines

davidvader:
Summary of changes:
- enhanced log line focus with (start, end) tuple
- added keypress tracking to main model to detect the shift key
- refactored log line focus, `<a>` -> `<button>` 
- added URL update control to Main.elm, rather than using HTML href
- enhance log link button with shift control (shift attempts to do range)
- enhance build URL fragment parsing to handle step, 1 line, and/or any range
- created Logs.elm for all logs view/helpers
- separated Builds and Build and moved them into `Pages/` module


davidvader:
one day Ill correctly name the PR the first time.... one day

PR #83, opened 2020-02-03
fix(nav): nav module and hooks page nav fix

davidvader:


src/elm/Main.elm
```diff
@@ -1542,6 +1444,7 @@ loadHooksPage model org repo maybePage maybePerPage =
     ( { model | page = Pages.Hooks org repo maybePage maybePerPage, hooks = loadingHooks, hookBuilds = Dict.empty }
     , Cmd.batch
         [ getHooks model org repo maybePage maybePerPage
+        , getCurrentUser model
```
davidvader: need current user to load favorites status

src/elm/Nav.elm
```diff
@@ -0,0 +1,156 @@
...
+                    , ( "-outline", True )
+                    ]
+                , onClick <| actions.restartBuild org repo buildNumber
+                , Util.testAttribute "restart-build"
+                ]
+                [ text "Restart Build"
+                ]
+
+        Pages.Hooks org repo _ _ ->
+            div [ class "buttons" ]
```
davidvader: this code was updated as well as moved, to fix some missed css cleanup from "-inverted" to "-outline"

PR #84, opened 2020-02-03
perf(html): lazy node rendering for pages root

davidvader:
Added lazy rendering to all pages.
https://guide.elm-lang.org/optimization/lazy.html

When wrapped in Html.lazyN, Elm will check the input used for the last frame render and compare it to the input for the current frame render. If the input and function have not changed, then the virtual DOM doesnt need to be updated and Elm will ignore that frame.

Best when used at the root of the application where the model is parsed and used as view input, so I added it to the Pages controller.

### DIFF
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -1194,7 +1194,7 @@ viewContent model =
 
         Pages.Build org repo buildNumber _ ->
             ( "Build #" ++ buildNumber ++ " - " ++ String.join "/" [ org, repo ]
-            , lazy4 Pages.Build.viewBuild model org repo buildActions
+            , lazy4 Pages.Build.viewBuild model org repo buildMsgs
             )
 
         Pages.Login ->
@@ -1391,11 +1391,6 @@ viewThemeToggle theme =
 
 
 -- HELPERS
-
-
-buildActions : Pages.Build.Actions Msg
-buildActions =
-    Pages.Build.Actions ClickStep UpdateUrl
 
 
 buildUrl : String -> List String -> List QueryParameter -> String
@@ -1821,6 +1816,13 @@ clickHook model org repo buildNumber =
         ( Dict.update ( org, repo, buildNumber ) (\_ -> Just buildInfo) model.hookBuilds
         , action
         )
+
+
+{-| buildMsgs : prepares the input record required for the Build page to route Msgs back to Main.elm
+-}
+buildMsgs : Pages.Build.Msgs Msg
+buildMsgs =
+    Pages.Build.Msgs ClickStep UpdateUrl
 
 
 {-| addReposMsgs : prepares the input record required for the AddRepos page to route Msgs back to Main.elm
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -1285,15 +1285,15 @@ viewNav : Model -> Html Msg
 viewNav model =
     nav [ class "navigation", attribute "aria-label" "Navigation" ]
         [ Crumbs.view model.page
-        , navButton model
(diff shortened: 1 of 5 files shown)


src/elm/Main.elm
```diff
@@ -1141,12 +1141,12 @@ viewContent model =
...
         Pages.Overview ->
             ( "Overview"
-            , Pages.Home.view model.user ToggleFavorite
+            , lazy2 Pages.Home.view model.user homeMsgs
             )
 
         Pages.AddRepositories ->
             ( "Add Repositories"
-            , Pages.AddRepos.view model addReposMsgs
+            , lazy2 Pages.AddRepos.view model addReposMsgs
```
davidvader: for example, since we pass in a PartialModel to the AddRepos page, we will only trigger a DOM rerender when the `sourceRepos` field of the main model changes, even if say, the `theme` is updated

src/elm/Main.elm
```diff
@@ -1187,14 +1187,14 @@ viewContent model =
             ( String.join "/" [ org, repo ] ++ " builds" ++ page
             , div []
                 [ Pager.view model.builds.pager Pager.defaultLabels GotoPage
-                , Pages.Builds.view model.builds model.time org repo
+                , lazy4 Pages.Builds.view model.builds model.time org repo
```
davidvader: although `time` in the main model is updated every few seconds via the `subscriptions` I still added `lazy` here for consistency and potential performance updates. (AKA, it couldnt hurt, right?)

src/elm/Main.elm
```diff
@@ -1818,13 +1818,34 @@ clickHook model org repo buildNumber =
         )
 
 
+{-| homeMsgs : prepares the input record required for the Home page to route Msgs back to Main.elm
```
davidvader: added some more Msgs helpers for code readability and consistency

src/elm/Main.elm
wass3r: we _are_ passing the whole model along here, so i don't think this would gain much as _lazy_ would evaluate the whole thing.. when we're passing fields of a model as we do above with `model.user`, i think we're getting the expected benefit. `AddRepos` accepts any record that at least has `user`, `sourceRepos`, and `filters` fields (see https://github.com/go-vela/ui/blob/master/src/elm/Pages/AddRepos.elm#L71-L73) and of course our model as a whole fits that criteria.

src/elm/Main.elm
```diff
@@ -1187,14 +1187,14 @@ viewContent model =
...
-                , Pages.Builds.view model.builds model.time org repo
+                , lazy4 Pages.Builds.view model.builds model.time org repo
                 , Pager.view model.builds.pager Pager.defaultLabels GotoPage
                 ]
             )
 
         Pages.Build org repo buildNumber _ ->
             ( "Build #" ++ buildNumber ++ " - " ++ String.join "/" [ org, repo ]
-            , Pages.Build.viewBuild model.time org repo model.build model.steps model.logs ClickStep UpdateUrl model.shift
+            , lazy4 Pages.Build.viewBuild model org repo buildMsgs
```
wass3r: I think this suffers from the same issue mentioned above with passing along the whole model.

src/elm/Main.elm
davidvader: just pushed some changes!

src/elm/Main.elm
davidvader: just pushed a couple changes. I thought `{ a | user = user }` was a _subset_ of the record given, not the entire thing. Should have known that tho!

davidvader:
conflictSSSSSSSSSSS! resolved :)

PR #89, opened 2020-02-04
fix(proxy): allow cli to fully work with ui url

wass3r:


davidvader:
I could use some more context :)

wass3r:
talk, we will

PR #98, opened 2020-02-05
fix(history): fix html/css causing focus outline and click bugs

davidvader:


davidvader:
Screen Shot 2020-02-05 at 12 13 11 PM

Screen Shot 2020-02-05 at 12 13 36 PM

src/scss/_main.scss
```diff
@@ -644,39 +656,39 @@ nav {
...
   &:before {
     content: '#';
   }
 }
 
-.build-history .recent-build .-event {
+.recent-build-tooltip .info {
+  padding: 0.2em 0.6em;
+
+  font-size: 14px;
```
wass3r: should refrain from using fixed font size.

i'm planning to clean that up in a few places. in this case `0.9rem` might do the trick.

src/elm/Pages/Build.elm
```diff
@@ -355,22 +371,32 @@ viewRecentBuild now timezone org repo buildNumber build idx =
...
+    tooltip is visible when the recent build link is focused or hovered
+
+-}
+recentBuildTooltip : Posix -> Zone -> Build -> Html msg
+recentBuildTooltip now timezone build =
+    div [ class "recent-build-tooltip", Util.testAttribute "build-history-tooltip" ]
+        [ div [ class "info" ]
+            [ div [ class "line", class "header" ]
+                [ span [ class "number" ] [ text <| String.fromInt build.number ]
+                , span [ class "event" ] [ text build.event ]
```
wass3r: i see `.event` just applies italic styling. as you want to stress this keyword visually, why not use `<em>` to call it out.

src/elm/Pages/Build.elm
```diff
@@ -355,22 +371,32 @@ viewRecentBuild now timezone org repo buildNumber build idx =
...
                 ]
+            , div [ class "line" ] [ span [ class "label" ] [ text "started:" ], span [ class "content" ] [ text <| Util.dateToHumanReadable timezone build.started ] ]
```
wass3r: looks like there is no `.content` css class, can probably drop the class.

i think we can also drop `.recent-build-tooltip .label` completely from the stylesheet and subsequently drop it from assignment here.

src/scss/_main.scss
```diff
@@ -644,39 +656,39 @@ nav {
...
+}
+
+.recent-build-tooltip .label {
```
wass3r: as mentioned above, i don't think this is necessary.

src/scss/_main.scss
```diff
@@ -644,39 +656,39 @@ nav {
...
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+}
+
+.recent-build-tooltip .event {
```
wass3r: as mentioned above, use `<em>` instead and drop this declaration

src/elm/Pages/Build.elm
```diff
@@ -355,22 +371,32 @@ viewRecentBuild now timezone org repo buildNumber build idx =
...
+{-| recentBuildTooltip : takes time info and build and renders tooltip for viewing recent build info
+
+    tooltip is visible when the recent build link is focused or hovered
+
+-}
+recentBuildTooltip : Posix -> Zone -> Build -> Html msg
+recentBuildTooltip now timezone build =
+    div [ class "recent-build-tooltip", Util.testAttribute "build-history-tooltip" ]
+        [ div [ class "info" ]
+            [ div [ class "line", class "header" ]
```
wass3r: i believe `.header` is unecessary here, as `.line` does the same thing

davidvader:
@wass3r change requests implemented 👍

davidvader:
@jbrockopp requesting one more fresh review, thank you for the feedback 
see
https://github.com/go-vela/ui/pull/98/commits/ce826eb00f1118766eb5237cd62bee11e7000149

PR #108, opened 2020-02-17
fix(help): accessibility tweaks for contextual help dropdown

davidvader:


src/elm/Help/Help.elm
```diff
@@ -50,13 +50,16 @@ view : Args msg -> Html msg
 view args =
     li
         [ id "contextual-help"
+        , attribute "aria-label" "toggle contextual help for this page"
```
davidvader: needed to promote this element to be tab focusable, because we are removing the tab focus from the button itself, due to the manual click handler that circumvents the built-in details click event

src/elm/Help/Help.elm
```diff
@@ -177,7 +180,7 @@ row content copy =
             Just copyMsg ->
                 copyButton
                     [ Util.testAttribute "help-copy"
-                    , attribute "aria-label" "view cli command for this page"
+                    , attribute "aria-label" <| "copy " ++ content ++ " to clipboard"
```
davidvader: added better aria label to copy button to give screen readers information as to what the button is doing AND copying

src/elm/Help/Help.elm
```diff
@@ -212,7 +215,11 @@ docsLink : Command -> Html msg
 docsLink command =
     case command.docs of
         Just docs ->
-            a [ class "cmd-link", href <| cliDocsUrl docs ]
+            a
+                [ class "cmd-link"
+                , href <| cliDocsUrl docs
+                , attribute "aria-label" <| "go to cli docs page for " ++ docs
```
davidvader: adding better aria label for docs link.
before, screen reader just read "link, docs"

PR #111, opened 2020-02-18
chore(deps): update dependencies

wass3r:


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "vela-ui",
+  "name": "vela-web-ui",
```
davidvader: any particular reasoning for the name change?

wass3r: yes, there is another package called `vela-ui` that results in this being shown at the top of our repo:

wat

PR #146, opened 2020-04-07
feat(secrets): pages for managing repo, org and shared secrets

davidvader:


src/scss/_main.scss
```diff
@@ -1354,6 +1343,160 @@ nav {
   }
 }
 
+.-m-r {
+  margin-right: 0.25rem;
+}
+
+.-m-l {
```
wass3r, line 1350: i think we were trying to stay away from using stand-alone modifier classes. they should always be chained to a box selector. in reviewing where this is used, we could utilize already established mechanisms by utlizing `.form-controls` and `.form-control` classes, though they might need to be extended?

src/scss/_main.scss
```diff
@@ -1354,6 +1343,160 @@ nav {
...
+  border-color: var(--color-primary);
+  position: absolute;
+  left: -1.5rem;
+}
+
+h4.-no-pad {
+  margin: 0;
+}
+
+h4.-m-t {
```
wass3r, line 1436: making note to chat about this one..

PR #150, opened 2020-04-24
feat(table): reusable table component for secrets/hooks/etc

davidvader:


davidvader:
do not merge.... VSCode auto formatted that scss file. Fixing!

src/scss/_main.scss
```diff
@@ -1313,6 +1345,34 @@ nav {
...
+  font-weight: 300;
+  margin-left: 1rem;
+}
+
+.field-header {
+  margin-top: 1rem;
+  margin-bottom: 0.25rem;
+}
+
+.-code {
```
wass3r, line 1372: as a modifier, this should have a element it's attached to. I don't actually see this being used. If it was, would wrapping the content in `<code>` be sufficient?

src/elm/Table/Table.elm
```diff
@@ -0,0 +1,135 @@
...
+customCell : Html msg -> Html.Attribute msg -> Html msg
+customCell element cls =
+    div [ class "cell", cls ]
+        [ span [] [ element ] ]
+
+
+{-| arrayCell : takes string array and renders cell
+-}
+arrayCell : List String -> String -> Html msg
+arrayCell images default =
```
wass3r, line 128: `images` seems weird here as a word for what looks to be textual content. what about `contents`?  and then rename the singular to `content`?

davidvader: doh! that was left in from working on the secrets table, my mistake

src/scss/_main.scss
davidvader: agreed! ill update

davidvader: oh that must have slipped in while dev'ing, its not even used. good catch, thank you!

src/elm/Table/Table.elm
davidvader: @jbrockopp dismissed your review to fix some copy pasta mistakes from ripping from my secrets branch

PR #165, opened 2020-05-26
fix(favicon): refresh the favicon even when the tab is hidden

davidvader:


src/elm/Main.elm
```diff
@@ -1332,6 +1346,24 @@ refreshPage model _ =
...
+refreshPageHidden : Model -> RefreshData -> Cmd Msg
+refreshPageHidden model _ =
+    let
+        page =
+            model.page
+    in
+    case page of
+        Pages.Build org repo buildNumber _ ->
+            Cmd.batch
+                [ refreshBuild model org repo buildNumber
```
davidvader, line 1360: when the page is hidden, we only need build status for the favicon.
this is to limit the load put on the server, while getting the functionality we need across inactive tabs.

PR #166, opened 2020-05-26
fix(refresh): refresh page when returning to an inactive tab

davidvader:


src/elm/Main.elm
```diff
@@ -1155,7 +1155,16 @@ update msg model =
             ( m, Cmd.none )
 
         VisibilityChanged visibility ->
-            ( { model | visibility = visibility, shift = False }, Cmd.none )
+            let
+                cmd =
+                    case visibility of
```
davidvader, line 1160: if _returning_ to a tab (setting it to visible) refresh the page we are viewing!

PR #167, opened 2020-05-27
fix(build): fix bug with focusing step when step is opened

davidvader:


davidvader:
> can we add some tests? probably:
> 
> 1. click step, confirm hash
> 2. click step, click another step, confirm hash
> 3. click step, click log line, click step again, confirm hash
> 
> to start with, plus any you can think of

done and done! added those tests, as well as a few other cases like visiting the page with a fragment then clicking a step, etc

cypress/integration/steps.spec.js
```diff
@@ -45,15 +45,61 @@ context('Steps', () => {
       // all test logs have a '$' encoded in the source
       cy.get('@logs').children().should('contain', '$');
     });
-    it('clicking steps should show/hide logs', () => {
+
+    it('logs should be hidden', () => {
```
davidvader, line 49: just cleaning up some tests/contexts here, while I was at it

cypress/integration/steps.spec.js
```diff
@@ -85,7 +131,6 @@ context('Steps', () => {
             cy.get('[data-test=log-line-num-2]').as('otherLineNumber');
           });
           cy.get('@otherLineNumber').click({ force: true });
```
davidvader, line 133: the bug was actually letting this bogus test pass. OOPS!

cypress/integration/steps.spec.js
```diff
@@ -85,7 +131,6 @@ context('Steps', () => {
...
-          cy.get('@stepHeaders').click({ force: true, multiple: true });
```
davidvader, line 88: the bug was actually letting this bogus test pass. OOPS!

PR #169, opened 2020-06-02
fix(help): --step in help window should be --step-number

meinenec:
Ran into an issue copying and pasting the command `vela view step --org foo --repo bar --build x --step x`

Changed to be step-number. Let me know if I missed something :D

### DIFF
--- a/src/elm/Help/Commands.elm
+++ b/src/elm/Help/Commands.elm
@@ -243,7 +243,7 @@ listSteps org repo buildNumber =
 {-| viewStep : returns cli command for viewing a step
 
     eg.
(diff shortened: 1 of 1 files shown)


davidvader:
Awesome thank you @meinenec !
My only hesitation.... `--build-number` was updated to accept `--build`. Should the same be said for `--step` ?
I will approve this PR, if @kneal or @jbrockopp approve first :)

meinenec:
> Awesome thank you @meinenec !
> My only hesitation.... `--build-number` was updated to accept `--build`. Should the same be said for `--step` ?
> I will approve this PR, if @kneal or @jbrockopp approve first :)

we could make the change there as well 😄

PR #170, opened 2020-06-04
fix(crumb): fix secrets crumbs display and add tests

davidvader:


src/elm/Crumbs.elm
```diff
@@ -241,7 +242,7 @@ toPath page =
                             ( org, Just <| Pages.OrgSecrets engine org Nothing Nothing )
 
                         nameCrumb =
-                            ( name, Nothing )
+                            ( Maybe.withDefault name <| percentDecode name, Nothing )
```
davidvader, line 245: the need to decode a team or name is to go alongside #171 that will encode urls for names that include special characters

cypress/integration/crumbs.spec.js
```diff
@@ -61,4 +61,73 @@ context('Crumbs', () => {
...
+        'fixture:secret_shared.json',
+      ).as('secret');
+      cy.login('/-/secrets/native/shared/github/octo%2Fcat/docker%2Fpassword');
+    });
+    it('should show appropriate secrets crumbs', () => {
+      cy.get('[data-test=crumb-secrets]').should('exist');
+      cy.get('[data-test=crumb-native]').should('exist');
+      cy.get('[data-test=crumb-shared]').should('exist');
+      cy.get('[data-test=crumb-github]').should('exist');
+      cy.get('[data-test="crumb-octo/cat"]').should('exist');
```
davidvader, line 115: special characters within test tags require double quotes

PR #171, opened 2020-06-04
fix(secrets): encode url when clicking on secret containing special characters

davidvader:


src/elm/Pages/Secrets/View.elm
```diff
@@ -157,16 +158,23 @@ renderSecret type_ secret =
 -}
 updateSecretHref : SecretType -> Secret -> Html.Attribute msg
 updateSecretHref type_ secret =
+    let
+        encodedTeam =
+            percentEncode secret.team
```
davidvader, line 163: added encode for team, because git allows teams containing `/`

PR #173, opened 2020-06-10
chore(deps): Update Makefile and bump dependencies

wass3r:
* Update Makefile with comments
* Add `bump-deps` targets
* Add `bump-deps-test` target
* Refactor `clean` target
* Update project dependencies
* Add `help` default target, so just running `make` or `make help` will show:
```
restart           Restart the local docker-compose stack
up                Create the local docker-compose stack
up-prod           Create the local docker-compose stack (prod)
down              Destroy the local docker-compose stack
pull              Pull all images for the local docker-compose stack
compose-up        Build and create containers for local docker-compose stack
compose-down      Destroy all containers for local docker-compose stack
clean             Clean local NPM and Elm dependencies
build             Build the UI in development mode
build-prod        Build the UI in production mode
test              Test the Elm source code
test-cypress      Run Cypress tests
format-validate   Validate Elm source code formatting
format            Format the Elm source code
bump-deps-npm     Bump NPM dependencies
bump-deps-elm     Bump Elm dependencies
bump-deps         Bump NPM and Elm dependencies
bump-deps-test    Bump dependencies and run Cypress tests
```


Makefile
```diff
@@ -58,3 +80,21 @@ test:
 .PHONY: format
 format:
 	@elm-format --yes src/ tests/
+
+.PHONY: bump-deps
+bump-deps: clean
```
jbrockopp, line 85: Thoughts on `upgrade` instead of `bump-deps`?

It's a little shorter and easier to remember, but it could have an unintended side effect of someone misinterpreting what the purpose for the command is 🤷‍♂️ 

I bring it up because I'm currently working on something similar and want to remain consistent 👍

wass3r: yea, that's what i had before and then i switched for brevity. i'm fine either way though.

jbrockopp: Hmmm... maybe for now we stick with `bump-deps` and consider `upgrade` later.

I was using `update` and `upgrade` terminology from an adoption of Homebrew.

But that might not be the best decision due to those who may be unfamiliar 🤷‍♂️

wass3r: only reason i'm hesitant is because there might be a use for a target like that that does more than upgrade dependencies. but it could still work.

jbrockopp: Sure 👍  

Unless others feel strongly, I'm good to use `bump` over `upgrade`.

Tell you what, I'll leave it up to you and then follow suit with whatever you decide 😉

wass3r: I think I’ll leave it as is and reserve just bump for potential future usage.

PR #179, opened 2020-07-08
fix(secrets): adding responsive table to secrets

davidvader:
will apply these changes to hooks after merging


src/scss/_table.scss
```diff
@@ -0,0 +1,319 @@
...
+// style below this should be removed when integrating new table with the rest of the application
+.table {
+    width: 100%;
+    margin: 1rem 0;
+    padding-bottom: 0.5rem;
+
+    font-size: 1rem;
+
+    background: var(--color-bg-dark);
+}
```
davidvader, line 124: this style is being move from _main.scss
most of the code below this section will be removed when adapting the hooks page to use the new table design

src/elm/Table.elm
```diff
@@ -0,0 +1,108 @@
...
+            [ div []
+                [ text label
+                , Maybe.withDefault (text "") headerElement
+                ]
+            ]
+        , thead [] [ tr [] <| List.map (\col -> th [ scope "col" ] [ text <| String.Extra.toTitleCase col ]) columns ]
+        , footer noRows numRows
+        , tbody [] <|
+            if List.length rows > 0 then
+                List.map (\row_ -> row_.display row_.data) rows
```
davidvader, line 93: most of the work is offset to the component utilizing the table

src/elm/Table.elm
```diff
@@ -0,0 +1,108 @@
...
+        [ Html.caption []
+            [ div []
+                [ text label
+                , Maybe.withDefault (text "") headerElement
+                ]
+            ]
+        , thead [] [ tr [] <| List.map (\col -> th [ scope "col" ] [ text <| String.Extra.toTitleCase col ]) columns ]
+        , footer noRows numRows
+        , tbody [] <|
+            if List.length rows > 0 then
```
wass3r, line 92: i don't think you need this if/else? if `rows` is 0 it will just produce an empty list anyway in `List.map`, wouldn't it?

src/scss/_table.scss
```diff
@@ -0,0 +1,320 @@
...
+
+.table .headers .first-cell {
+  width: 2rem;
+}
+
+.filler-cell {
+  width: 6ch;
+}
+
+.table .headers .-label {
```
wass3r, line 215: is `.-label` attached to another class or is it a generic utility class?

src/scss/_table.scss
```diff
@@ -0,0 +1,320 @@
...
+.table .info.-success {
+  border-left-color: var(--color-green);
+}
+
+.table .info.-failure,
+.table .info.-error {
+  border-left-color: var(--color-red);
+}
+
+.table .element span.-m-r,
```
wass3r, line 282: i think we're using `.-m-r` beyond the scope of this PR, so we'll address changing it to utility class in a separate PR probably.

should that `span` in this case have it's own class?

src/elm/Pages/Secrets/View.elm
```diff
@@ -142,18 +150,69 @@ tableHeaders =
...
-        , Table.Table.arrayCell secret.images "all images"
-        , Table.Table.cell (Util.boolToYesNo secret.allowCommand) <| class ""
-
-        -- TODO: change linked name to edit button, when table is fixed
-        -- , Table.Table.customCell ( a [ class "button", class "-outline", updateSecretHref type_ secret ] [ text "edit" ]) <| class ""
+    tr [ Util.testAttribute <| "secrets-row" ]
+        [ td
+            [ attribute "data-label" "name"
+            , scope "row"
+            , class "-line-break"
```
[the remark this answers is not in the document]
davidvader: agreed, will convert this to a utility class.
I believe I originally had a `.row` class but have since removed it

src/elm/Table.elm
davidvader: agreed, good catch. was originally using that if/else to render the "no rows" info but have since moved that to `footer`.
good catch!

src/scss/_table.scss
davidvader: anything related to `.table` will be removed with hooks, consider these all dead

src/scss/_table.scss
davidvader: see comment above, anything below/attached to `.table` will be gutted and removed with hooks refactor

src/elm/Table.elm
davidvader: fixed 👍

src/elm/Pages/Secrets/View.elm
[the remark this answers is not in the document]
davidvader: fixed 👍 
along with 2 incorrect modifiers for `break-all` and `overflow-auto`

src/scss/_table.scss
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  .table-base td:last-child {
+    border-bottom: 0;
+  }
+}
+
+// end table-base
+
+// style below this should be removed when integrating new table with the rest of the application
```
davidvader, line 112: everything below this is null and void, and will be removed with the next PR integrating Table with Hooks

PR #180, opened 2020-07-08
fix: correct issue base URL

tmapes:
Noticed this when clicking one of the 'upvote feature' button in the CLI thing in the top right corner.


jbrockopp:
@tmapes thanks for your contribution! 🎉

PR #184, opened 2020-08-03
feat: adding a skip step and icon to build

jenmartTarget:
image

image


jenmartTarget:
New icon "skipped" added. Changes made in SvgBuilder.elm. 

Updated steps.spec.js to reflect new icon addition. Now checks if test is valid, text appears and doesn't appear when requested. 
Added in new file "steps_5_skipped_step.json" to be used in steps.spec.js instead of using "steps_5.json".

src/elm/SvgBuilder.elm
```diff
@@ -119,8 +119,10 @@ buildSuccess =
         ]
 
 
-{-| buildFailure : produces svg icon for build status - failure
--}
+
+{- build new icon and case like above for skidpped icon. "kill is skidpped" -}
+
+
```
davidvader, line 125: should this change be reverted?

src/scss/_main.scss
```diff
@@ -971,6 +971,7 @@ nav {
 
   &.-failure,
   &.-error {
+    //here
```
davidvader, line 974: guessing this comment was left in by mistake?
```suggestion
```

src/scss/_main.scss
```diff
@@ -982,6 +983,11 @@ nav {
   .-inner {
     fill: var(--color-primary);
   }
+
+  &.-skip {
+    //here
```
davidvader, line 988: same with this?
```suggestion
```

src/elm/SvgBuilder.elm
```diff
@@ -261,6 +263,66 @@ stepFailure =
         ]
 
 
+{-| stepSkipped : produces svg icon for step status - killed
+-}
+{-| Note: killed and skipped means the same thing.  -}
```
davidvader, line 268:
```suggestion
     Note: killed and skipped means the same thing.  -}
```
Kind of nit-picky but these comment brackets are redundant

src/scss/_main.scss
jenmartTarget: Removed.

src/elm/SvgBuilder.elm
jenmartTarget: Reverted.

cypress/integration/steps.spec.js
```diff
@@ -295,10 +301,10 @@ context('Steps', () => {
       cy.get('[data-test=step]').first().should('not.have.class', '-last');
     });
 
-    it('last step should not contain error', () => {
+    it('last step should not contain killed/skipped', () => {
       cy.get('[data-test=step]').last().as('echoStep');
       cy.get('@echoStep').should('be.visible').click({ force: true });
-      cy.get('@echoStep').should('not.contain', 'error:');
+      cy.get('@echoStep').should('not.contain', 'step was killed');
```
wass3r, line 307: i see we removed the previous check of to make sure it doesn't contain an error. I think we want to keep that check and add this new check on top of it, because it should be neither. you'll want to update the description of the test then as well.

src/scss/_main.scss
wass3r: unless i'm mistaken i still see this in the latest changeset.

src/scss/_main.scss
wass3r: unless i'm mistaken i still see this in the latest changeset.

cypress/integration/steps.spec.js
jenmartTarget: Added.

PR #185, opened 2020-08-04
fix(hooks): adapt hooks to use new table component

davidvader:


cypress/integration/hooks.spec.js
```diff
@@ -68,65 +68,19 @@ context('Hooks', () => {
...
-            it('should be running', () => {
-              cy.get('@build')
-                .get('.hook-build-status')
-                .should('have.class', '-running');
-            });
-            it('should show duration', () => {
-              cy.get('@build').contains('duration');
-            });
-          });
-        });
```
davidvader, line 129: removed build fetching, therefore removing the tests

PR #190, opened 2020-08-13
enhance(skip): change kill to skip

wass3r:
Match text with skip icon previously introduced.

### DIFF
--- a/src/scss/_themes.scss
+++ b/src/scss/_themes.scss
@@ -105,7 +105,8 @@ body.theme-light {
     border-bottom: 1px solid var(--color-bg-dark);
   }
 
-  .step-error {
+  .step-error,
+  .step-skipped {
     color: var(--color-red);
   }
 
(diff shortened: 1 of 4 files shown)


src/scss/_themes.scss
```diff
@@ -105,7 +105,8 @@ body.theme-light {
     border-bottom: 1px solid var(--color-bg-dark);
   }
 
-  .step-error {
+  .step-error,
+  .step-skipped {
```
jbrockopp, line 109: Forgive me inexperience with this 🐼 

Looking at this, it almost suggests to me that the `.step-skipped` class will apply the color red?

Am I reading or understanding this correctly?

wass3r: no worries. good question. yes, for the text color. we could keep it white/gray 🤔

jbrockopp: Ahhhhhh this would be for the text color in the log output 👍 

For some reason when I was reading this it lead me to believe that color would be applied across the board 😛

wass3r: i think i'm ok with the "warning" text color for now, but we discuss further beyond this PR 👍

PR #193, opened 2020-08-18
feat(logs): add actions for logs

davidvader:
Improving the experience for interacting with logs.
Adding buttons for 
- jumping to top/bottom
- download logs
- tracking new logs per step
- auto expand build steps as they run


src/elm/Pages/Build/Update.elm
```diff
@@ -64,43 +77,143 @@ update model msg getBuildStepLogs =
...
+
+
+{-| isViewingStep : takes steps and step number and returns the step viewing state
+-}
+isViewingStep : WebData Steps -> StepNumber -> Bool
+isViewingStep steps stepNumber =
+    steps
+        |> RemoteData.withDefault []
+        |> List.filter (\step -> String.fromInt step.number == stepNumber)
+        |> List.map (\step -> step.viewing)
```
wass3r, line 180:
```suggestion
        |> List.map .viewing
```
minor tweak

src/scss/_main.scss
```diff
@@ -1333,3 +1356,19 @@ nav {
 .overflow-auto {
   overflow: auto;
 }
+
+.flex {
```
wass3r, line 1360: these are starting to accumulate. just a note that we should consider separating these generic utility classes out soon.

src/scss/_tooltips.scss
```diff
@@ -0,0 +1,99 @@
...
+  margin: 0.75rem 0;
+  padding: 0.2rem 0.4rem;
+
+  font-size: 0.8rem;
+  line-height: 1.2;
+
+  content: attr(data-tooltip);
+}
+
+[data-tooltip]:hover:before,
```
wass3r, line 45: while this is not wrong, let's use the proper `::` for pseudo-selectors. it'll also be consistent with other instances.

src/scss/_main.scss
```diff
@@ -419,18 +419,37 @@ nav {
...
 .build .error .message {
   margin-left: 0.2em;
 }
 
+.log-actions {
+  display: flex;
+  justify-content: flex-end;
+  margin-top: 0.2rem;
+  margin-right: 1rem;
+  background: linear-gradient(
```
wass3r, line 445: does this need to be a gradient? i think simple background color is fine here?

src/scss/_main.scss
```diff
@@ -419,18 +419,37 @@ nav {
...
 }
 
 .build .error .message {
   margin-left: 0.2em;
 }
 
+.log-actions {
+  display: flex;
+  justify-content: flex-end;
+  margin-top: 0.2rem;
```
wass3r, line 443: toggling this on/off doesn't change much. can we drop it?

src/scss/_main.scss
davidvader: good point, and there will be some more in PRs I have in flight, I'll consider this sooner rather than later in a smaller PR :)

src/scss/_main.scss
davidvader: thats actually the flowline that connects step icons on the left side. I've renamed the class to `flowline-left` and adapted it to the new log-actions, removing it from this class ^ Feel free to open another comment if you want but that linear gradient is how I get it to work well with icon padding and overflow-hidden

src/scss/_main.scss
davidvader: Great catch. Done! 👍

src/scss/_main.scss
wass3r: yup, you're right, my fault. i think what happened is that i toggled on/off and thought it was doing the opposite - blocking the line out. nothing to see here :)

src/scss/_tooltips.scss
wass3r: to clarify:
`[data-tooltip]:hover:before` -> `[data-tooltip]:hover::before`

davidvader: got it..... another great catch. Done! 🦖

PR #194, opened 2020-08-24
enhance(overview): make entire repos on Overview clickable

davidvader:
Move the "View" repo functionality on Overview from a button to the entire repo item/row

### DIFF
--- a/cypress/integration/favorites.spec.js
+++ b/cypress/integration/favorites.spec.js
@@ -186,7 +186,7 @@ context('Favorites', () => {
               cy.route('PUT', '*api/v1/user*', 'fixture:favorites.json');
               cy.get('[data-test=star-toggle-cat-purr]').as('togglePurr');
               cy.get('@togglePurr').click();
-              cy.get('[data-test=star-toggle-cat-purr]').should(
+              cy.get('a > [data-test=star-toggle-cat-purr]').should(
                 'not.be.visible',
               );
             });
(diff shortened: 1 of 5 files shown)


davidvader:
> i think this change introduces a few things to consider:
> 
> 1. while the cypress test covers the functionality of the whole "button" now, we don't have any tests for the buttons within. you'll notice if you add them (or just try using them), that the buttons don't function as expected due to event bubbling.
> 2. related to the first point, besides being a bit awkward and causing a11y issues, it's actually invalid for `<a>` to contain interactive content (ie. other `<a>` or `<button>`, etc) according to the HTML spec.
> 3. with the above in mind, one option is to consider dropping the buttons (but leaving the favoriting ability, shifting it outside of the `<a>` element), so the way to access the lost "button" functionality moves to where you are in the context of a repo (ie. viewing a repo) - we have them there already.
> 4. lastly, i think it would be helpful to have some visual cues (on top of changing the pointer) to signal to the user that the whole "row" is interactive.

Absolutely these are extremely important callouts. (I expected this I should have made the PR a draft to prompt discussion much like the other proposals). I'll take what you said into consideration and work up another pass to this 👍 
I plan to rework the other action buttons outside of the interactive row, implement more tests and double check accessibility requirements. Thank you! @wass3r

davidvader:
In progress of implementing this through a series of navigation/functionality improvements. 
Closing in lieu of discussion on 8/27/2020.

PR #199, opened 2020-09-04
fix(css): ignore ansi stylesheet during purgecss

davidvader:
ansi styles are applied during runtime, therefore purgecss is considering the ansi stylesheet as "unused" and is consequently scrubbing it from the production parcel build.
this commit adds ignore flags to _ansi.scss


wass3r (approved):

kneal (approved):
LGTM 🐬

PR #200, opened 2020-09-11
fix(css): give new hover tooltips breathing room

davidvader:
fixes arrow position and adds a small amount of margin to tooltip content


src/scss/_tooltips.scss
```diff
@@ -88,9 +87,9 @@
 .tooltip-left:hover::after,
 .tooltip-left:focus::before,
 .tooltip-left:focus::after {
-  -webkit-transform: translateX(-0.8rem);
-  -moz-transform: translateX(-0.8rem);
-  transform: translateX(-0.8rem);
+  -webkit-transform: translateX(-0.6rem);
```
wass3r, line 90: we shouldn't have to include vendor prefixes since we use autoprefixes (see https://github.com/go-vela/ui/blob/master/.postcssrc.js)

wass3r (changes requested):

src/scss/_tooltips.scss
davidvader: oh good point I knew that, copy pasta. fixed!

wass3r (approved):

kneal (approved):
LGTM 🐬

PR #197, opened 2020-08-27
enhance(logs): scrollable/windowed logs and actions sidebar

davidvader:
Proposing an enhancement to step logs to wrap them in scrollable/resizable window.

Depends on #193 

### UX Improvement

- Reduce overall screen size
- Improved scrolling performance
- More interactive/modern feel

### Performance

The following performance summaries were gathered and summarized over 20~ runs using Google Chrome's performance profiler, auto expand/collapsing 5 step logs each with 500+ lines.

##### Before

ui_expand_before

##### After

ui_expand_after

#### Conclusion

Reducing screen clutter and browser workload, time spent scripting/dynamically generating content is converted to idle time. However low sample size may be misleading. 

### Example

Screen Shot 2020-08-27 at 2 37 28 PM

PR #202, opened 2020-09-14
fix(css): step-error and step-killed padding

davidvader:
removing unused classes
```css
.step-error .message,
.step-skipped .message
```
and moving padding to step-killed
(in preparation for #197 )

### DIFF
--- a/src/elm/Pages/Build/View.elm
+++ b/src/elm/Pages/Build/View.elm
@@ -651,14 +651,12 @@ viewError build =
         Vela.Error ->
             div [ class "error", Util.testAttribute "build-error" ]
                 [ span [] [ text "error:" ]
-                , span [ class "message" ]
-                    [ text <|
-                        if String.isEmpty build.error then
-                            "no error msg"
+                , text <|
+                    if String.isEmpty build.error then
+                        "no error msg"
 
-                        else
-                            build.error
-                    ]
+                    else
+                        build.error
                 ]
 
         _ ->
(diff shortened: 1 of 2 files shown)


jbrockopp (approved):
LGTM

src/elm/Pages/Build/View.elm
```diff
@@ -592,7 +592,7 @@ stepFollowButton stepNumber following =
 stepError : Step -> Html msg
 stepError step =
     div [ class "step-error", Util.testAttribute "step-error" ]
-        [ span [ class "label" ] [ text "error:" ]
+        [ span [] [ text "error:" ]
         , span [ class "message" ]
```
wass3r, line 596: can we get rid of this `span` then?

src/elm/Pages/Build/View.elm
```diff
@@ -637,7 +637,7 @@ viewError build =
     case build.status of
         Vela.Error ->
             div [ class "error", Util.testAttribute "build-error" ]
-                [ span [ class "label" ] [ text "error:" ]
+                [ span [] [ text "error:" ]
                 , span [ class "message" ]
```
wass3r, line 641: can we get rid of this `span` then?

wass3r (commented):
i don't believe https://github.com/go-vela/ui/blob/d5a4e8a526314682637e99082c672e6abc543453/src/scss/_main.scss#L436 is used either, though i only did a cursory check.

PR #204, opened 2020-09-22
perf(logs): moving base64 decode to log receiver

davidvader:
preparing for future log updates, moving base64 decode to when log is received, rather than on view/render.
this should improve performance greatly and goes in hand with additional incoming enhancements

### DIFF
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -735,7 +735,7 @@ update msg model =
 
         StepLogResponse stepNumber logFocus refresh response ->
             case response of
-                Ok ( _, log ) ->
+                Ok ( _, incomingLog ) ->
                     let
                         following =
                             model.followingStep /= 0
@@ -763,8 +763,11 @@ update msg model =
 
                             else
                                 Cmd.none
+
+                        decodedLog =
+                            { incomingLog | data = Util.base64Decode incomingLog.data }
                     in
-                    ( updateLogs { model | steps = steps } log
+                    ( updateLogs { model | steps = steps } decodedLog
                     , cmd
                     )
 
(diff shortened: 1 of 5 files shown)


src/elm/Pages/Build/Logs.elm
```diff
@@ -321,18 +320,6 @@ logFocusExists steps =
         /= ( Nothing, Nothing )
 
 
-{-| base64Decode : returns a string from a Maybe Log and decodes it from base64
```
davidvader, line 324: moving this to Util.elm

src/elm/Pages/Build/View.elm
```diff
@@ -305,14 +304,11 @@ viewLogs org repo buildNumber step logs follow shiftDown =
 viewLogLines : Org -> Repo -> BuildNumber -> StepNumber -> LogFocus -> Maybe (WebData Log) -> Int -> Bool -> Html Msg
 viewLogLines org repo buildNumber stepNumber logFocus maybeLog following shiftDown =
     let
-        log =
+        decodedLog =
```
davidvader, line 307: log is now decoded in Main.elm, no need to decode it in view/render

src/elm/Main.elm
```diff
@@ -763,8 +763,11 @@ update msg model =
 
                             else
                                 Cmd.none
+
+                        decodedLog =
```
davidvader, line 767: this is the meat of the enhancement, decode the log and overwrite the data, no need to do this in view/render and we shouldnt

wass3r (approved):

jbrockopp (approved):
LGTM

PR #205, opened 2020-09-22
perf(logs): implement javascript port for base64 decoding

davidvader:
Elm-only solutions for decoding base64 leave a lot to be desired.
https://package.elm-lang.org/packages/danfishgold/base64-bytes/latest/
https://package.elm-lang.org/packages/truqu/elm-base64/latest/
https://package.elm-lang.org/packages/danfishgold/base64-bytes/latest/

Unfortunately the above libraries still do not handle large strings well, and choke at string verifications.

Ive implemented a javascript interop port for decoding base64 using `atob` which does not check for string validity (which we dont really care about in our case)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,66 @@
+name: 'CodeQL'
+
+on:
+  push:
+    branches: [master]
+  pull_request:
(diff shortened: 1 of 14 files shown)


davidvader:
closing due to discrepancies on ANSI output using this base64 method

PR #215, opened 2020-10-08
perf(logs): optimization, only base64 decode when logs change

davidvader:
optimizing base64 decode process a bit with some small tweaks:
- update to use more performant base64 library (javascript solution still WIP)
- use updateIf rather than setIf when overwriting logs
- only base64 decode when we actually update the logs in the model

saw several performance improvements
large-logs performance bottleneck has been more-or-less transferred to table/row renderanimationframe bloating

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -1,44 +1,42 @@
 {
-    "type": "application",
-    "source-directories": [
-        "src/elm"
-    ],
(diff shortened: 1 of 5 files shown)


jbrockopp (approved):
LGTM

wass3r (approved):

PR #218, opened 2020-10-12
fix(css): adding break-word to logs

davidvader:
Adding 
```css
word-break: break-word;
```

to fix a bug with single-word logs that extend past the screens viewport.

### Before (extends past screen width):
Screen Shot 2020-10-12 at 3 33 38 PM

### After:
Screen Shot 2020-10-12 at 3 33 59 PM

### DIFF
--- a/src/scss/_main.scss
+++ b/src/scss/_main.scss
@@ -1418,7 +1418,7 @@ nav {
   @include visually-hidden();
 }
 
-.log-line-word-break {
+.break-text {
   word-break: break-all;
   word-wrap: break-word;
 }
--- a/src/scss/_main.scss
+++ b/src/scss/_main.scss
@@ -1419,7 +1419,7 @@ nav {
 }
 
 .break-all {
-  word-break: break-word;
+  word-break: break-all;
 }
 
 .break-word {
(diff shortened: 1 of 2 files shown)


src/scss/_main.scss
```diff
@@ -1418,8 +1418,9 @@ nav {
   @include visually-hidden();
 }
 
-.break-all {
+.log-line-word-break {
```
wass3r, line 1421: i would propose using something more generic like `.break-text` maybe.

wass3r (changes requested):

src/scss/_main.scss
davidvader: sounds good, I was just leaning away from the generic stylizers based on previous conversation but Im definitely cool with it 👍

davidvader: done 🚀

wass3r: Ah, I don't recall that. I do remember we talked about eventually creating a utilities import that would have things like this in it. I can see this being useful elsewhere potentially. Thanks!

wass3r (approved):

src/scss/_main.scss
davidvader: agreed 👍  I'll jot that down in my notes to create a util css file to put all these random classes

wass3r (approved):

jbrockopp (approved):
LGTM

PR #216, opened 2020-10-09
feat(stages): implement stages view

davidvader:
Implements viewing a pipeline that contains staged steps.
Changes can be summarized with
- Small refactor to Build.View to reduce function parameter noisiness
- When steps contain stages, group them
- Render a label for each stage, except for reserved stages like init/clone

PR #220, opened 2020-10-12
fix(css): adding gray color var to match current release flowline color

davidvader:
Fixing the flowline color. In #216 the hardcoded color was removed but using `var(--color-bg-light)` is more difficult to see. This change adds a better gray for light/dark theme and uses it for the flowline.

See below.
*After* is almost identical to the hardcoded color in the current release (0.7.0)

### Before: (--color-bg-light)
Screen Shot 2020-10-12 at 4 05 45 PM
Screen Shot 2020-10-12 at 4 05 53 PM

### After: (--color-gray)
Screen Shot 2020-10-12 at 4 02 45 PM
Screen Shot 2020-10-12 at 4 02 36 PM

### DIFF
--- a/src/scss/_main.scss
+++ b/src/scss/_main.scss
@@ -749,7 +749,7 @@ nav {
   background: linear-gradient(
     90deg,
     hsla(0, 0, 0, 0) calc(3.25em - 1px),
-    var(--color-bg-light) calc(3.25em),
+    var(--color-gray) calc(3.25em),
     hsla(0, 0, 0, 0) calc(3.25em + 1px)
   );
 }
--- a/src/scss/_main.scss
+++ b/src/scss/_main.scss
@@ -749,7 +749,7 @@ nav {
   background: linear-gradient(
     90deg,
     hsla(0, 0, 0, 0) calc(3.25em - 1px),
-    var(--color-gray) calc(3.25em),
+    var(--color-flowline) calc(3.25em),
     hsla(0, 0, 0, 0) calc(3.25em + 1px)
   );
 }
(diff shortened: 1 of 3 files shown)


wass3r (changes requested):
the downside to this approach is that if you want to theme the flowline, you have to redefine `--color-gray`. we probably want to assign the new color to `--color-bg-light`. not sure what else it would affect though. otherwise, if it's really unique to the flowline, make a new variable with that color assigned.

davidvader:
> the downside to this approach is that if you want to theme the flowline, you have to redefine `--color-gray`. we probably want to assign the new color to `--color-bg-light`. not sure what else it would affect though. otherwise, if it's really unique to the flowline, make a new variable with that color assigned.

Wow haha I actually had this implementation using `var(--color-flowline)` with a specific color for it but for some reason went with a generic `--color-gray` because I didnt want to bloat our variables with non-color related color variables... if that makes sense 😓 

Secondly, the reason Im actually making this change was because `bg-light` doesnt create a 1:1 relationship between light and dark theme. 

aka, when trying to create a clear distinction between the background and the foreground, `bg-light` is fine for the light theme and too feint for the dark theme, but `coal-light` also doesnt work, hence the addition of `gray`.

davidvader:
created `--color-flowline` and assigned it the value `--color-gray` in both dark and light theme, TY @wass3r for the recommendation and the review, let me know if I can make any other updates!!

wass3r:
Thanks. Looks good. I think slate is kind of in the same boat. Should probably add color-line-focus or something. Line focus might not be slate in other themes, ie. assigning red to slate is more awkward than assigning red to line focus.

wass3r (approved):

wass3r:
Also, agree on bg-light comments. We may want to base it on depth instead.

jbrockopp (approved):
LGTM

PR #223, opened 2020-10-16
Adds timestamp to build preview

bmuenzenmeyer:
👋 Hello! Very open feedback 💬 , coaching ⚔️,  pivoting 👢 , team takeover 🤼 , or throwing this in the 🗑️. 

Adds timestamp from when a build was created to the build preview as a title attribute. This maintains the existing behavior of humanized, relative-to-now time, but on hover, reveals the full date and time in ~UTC~ the user's timezone. This tooltip is visible on the repo builds list and the build itself.

image

> Notice the date / time tooltip while my (hidden) cursor hovers over "yesterday"

Already figured out `elm-format` but I am sure there are more elements this change may need to manifest, such as tests.

~I also could not figure out yet how to make this time respect the user's timezone. I seek guidance on how to do that within Elm / Vela... I see that `Main.elm` appears to have this information in scope.~

Now the user's timezone is respected. The same build from the original screenshot, now reflecting when the build ran relative to my local time.

image

Resolves: https://github.com/go-vela/community/issues/58


davidvader:
Looks great! Thanks @bmuenzenmeyer.
If its possible, is there any chance I could convince you to add a test for this in
https://github.com/go-vela/ui/blob/70004ebdc6b71e81e24b51cfca79414377cab6dd/cypress/integration/builds.spec.js ?

bmuenzenmeyer:
I am happy to add tests @davidvader but wasn't quite sure if the UTC timestamp is acceptable over a user-timezone-adjusted value. I am unsure how do that yet.

davidvader:
Sure no problem! Its actually not so bad, since you've already noticed we store this information in our application's data model. We just need to do a couple things to get it passed into viewPreview and you can replace the Time.utc.

For some reason the title attribute is not working for me locally however, for chrome or firefox!

I am investigating.... 

(ive had trouble with this before, hence my usage of a custom attribute and content for tooltips but the overhead is undesirable so Id like to figure out why title isnt working)
https://github.com/go-vela/ui/blob/8508b8090f456b3b15fef1a3f747cbed111a0c28/src/elm/Pages/Build/View.elm#L594
https://github.com/go-vela/ui/blob/310e8289439d2771e1f9facca26806c4b0cf9a13/src/scss/_tooltips.scss#L39

davidvader:
As far as the timezone is concerned I can outline some steps in the meantime. We need to...

1. Add `zone` to the PartialModel for Builds. PartialModel's let us pass specific fields from the Model found in Main.elm on a need-basis.
https://github.com/go-vela/ui/blob/310e8289439d2771e1f9facca26806c4b0cf9a13/src/elm/Pages/Build/Model.elm#L34

2. Pass in `zone` to the PartialModel when rendering a single Build in Main.elm
https://github.com/go-vela/ui/blob/310e8289439d2771e1f9facca26806c4b0cf9a13/src/elm/Main.elm#L1659

3. Add `zone` as a function parameter for Pages.Builds.view and trickle that parameter down to viewPreview
(I have yet to refactor Builds.view to use the PartialModel, but plan to someday)
https://github.com/go-vela/ui/blob/310e8289439d2771e1f9facca26806c4b0cf9a13/src/elm/Main.elm#L1649
^ you will also need to update `lazy5` to be `lazy6` since we are adding another parameter

Once those are done... you should be able to add Zone as a function parameter to viewPreview like this:
```elm
viewPreview : Posix -> Zone -> Org -> Repo -> Build -> Html Msg
```

which should let you use it instead of Time.utc!

bmuenzenmeyer:
Thanks for these breadcrumbs - I'll take a look on Friday if that's not too late

bmuenzenmeyer:
The repo hooks page looks like another good place for this feature - as I was trying to validate the timestamp first in GHE (it doesn't go back far enough), and then in Vela itself - but the created date is humanized too

image

If this is accepted - I can take a look into this as a separate body of work

bmuenzenmeyer:
@davidvader thanks for the pointers - this is ready for final review

davidvader (approved):
nice work, LGTM! 🚀

wass3r (approved):

wass3r:
Thanks @bmuenzenmeyer . We'd be happy to take further applications of this where it makes sense 👍

PR #225, opened 2020-10-26
feat(build): add recent builds title

KellyMerrick:
* Indented `Recent Builds` will be inline with the default selected most recent build in the build history

image

### DIFF
--- a/src/elm/Pages/Build/View.elm
+++ b/src/elm/Pages/Build/View.elm
@@ -758,7 +758,7 @@ viewBuildHistory now timezone page org repo builds limit =
                 if List.length blds > 0 then
                     div [ class "build-history" ]
                         [ p [ class "build-history-title" ] [ text "Recent Builds" ]
-                        , ul [ class "-no-pad", Util.testAttribute "build-history" ] <|
+                        , ul [ Util.testAttribute "build-history" ] <|
                             List.indexedMap (viewRecentBuild now timezone org repo buildNumber) <|
                                 List.take limit blds
                         ]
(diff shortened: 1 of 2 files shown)


src/elm/Pages/Build/View.elm
```diff
@@ -755,9 +756,12 @@ viewBuildHistory now timezone page org repo builds limit =
         case builds of
             RemoteData.Success blds ->
                 if List.length blds > 0 then
-                    ul [ class "build-history", class "-no-pad", Util.testAttribute "build-history" ] <|
-                        List.indexedMap (viewRecentBuild now timezone org repo buildNumber) <|
-                            List.take limit blds
+                    div [ class "build-history" ]
+                        [ p [ class "build-history-title" ] [ text "Recent Builds" ]
+                        , ul [ class "-no-pad", Util.testAttribute "build-history" ] <|
```
davidvader, line 761: since we moved the `-no-pad` style to the `ul` and removed the `.build-history.-no-pad` class, can we remove it from the element?
```suggestion
                        , ul [ Util.testAttribute "build-history" ] <|
```

src/scss/_main.scss
```diff
@@ -581,16 +582,21 @@ nav {
...
+    display: flex;
+    flex-direction: row;
+    align-items: center;
     margin: 0;
     padding: 0;
   }
 }
 
+.build-history-title {
+  font-size: 80%;
```
davidvader, line 595: Would you mind changing 80% to 0.8em?
Effectively the same but we're trying to be consistent. (when we can, and here, we can!)
```suggestion
  font-size: 0.8em;
```

davidvader (dismissed):

davidvader (approved):
looks great 👍  thanks for making those changes!!!

jbrockopp (approved):
LGTM

PR #226, opened 2020-10-27
docs: update info for local development

jbrockopp:
Related to https://github.com/go-vela/server/pull/202 and https://github.com/go-vela/server/pull/204 and https://github.com/go-vela/worker/pull/122

It's been a long time since the local development docs have been updated so thought I'd give it a try.

For convenience, here are links to the old/new local development docs:

* [Old](https://github.com/go-vela/ui/blob/master/.github/DOCS.md)
* [New](https://github.com/go-vela/ui/blob/docs/local/DOCS.md)

Found within this PR are the following changes:

* moved the `DOCS.md` to the root of the repository so it's easier to find (previously was nested under `.github`)
* updated the `DOCS.md` content to be more focused on starting the Vela stack using `make`
* updated the `DOCS.md` with information on the different compose services
* added to the `DOCS.md`  instructions for running builds locally when the Vela stack is started
* updated the `CONTRIBUTING.md` to reference the `DOCS.md` to reduce duplication of docs
* updated the `docker-compose.yml` to use `'` for strings (previously, we were using `'` and `"` interchangeably)
* updated the `.gitignore` to exclude the `secrets.env` file needed for [go-vela/server](https://github.com/go-vela/server)
* added a new `SUPPORT.md` doc that contains useful information on getting support
* removed the `CHANGELOG.md` file as it was empty and was not used


DOCS.md
```diff
@@ -0,0 +1,227 @@
...
+
+**NOTE: Please review the [start section](#start) before moving forward.**
+
+In order to run a build in Vela, you'll need to add a repo to the locally running system.
+
+<details><summary>click to reveal content</summary>
+<p>
+
+1. Navigate to the `Source Repositories` page @ http://localhost:8888/account/source-repos
+  * For conveinence, you can reference our documentation to [learn how to enable a repo](https://go-vela.github.io/docs/usage/getting-started/enable_repo/).
```
wass3r, line 119:
```suggestion
  * For convenience, you can reference our documentation to [learn how to enable a repo](https://go-vela.github.io/docs/usage/getting-started/enable_repo/).
```

DOCS.md
```diff
@@ -0,0 +1,227 @@
...
+
+For more information, please see our [installation docs](https://go-vela.github.io/docs/install/).
+
+## Prerequisites
+
+This section covers the dependencies required to get the Vela application running locally.
+
+* [Docker](https://docs.docker.com/install/) - building block for local development
+* [Docker Compose](https://docs.docker.com/compose/install/) - start up local development
+* [Github OAuth Client](https://developer.github.com/apps/building-oauth-apps/creating-an-oauth-app/) - building block for local Development
```
wass3r, line 13: to stay consistent

```suggestion
* [Github OAuth Client](https://developer.github.com/apps/building-oauth-apps/creating-an-oauth-app/) - building block for local development
```

wass3r (approved):
Mega-thanks for the edits!

I think it's good as-is, but made a couple (super) minor suggestions.

To note, the document seems slightly more geared towards how to start the whole stack (which works well), but I would not do most UI development in this manner. It, arguably, should be a step in your local dev prior to pushing your work up, but for incremental changes this is a bit painful.

Using things like `npm run dev` to run the UI layer with a file watcher that hot-reloads the browser as you are making changes is pretty standard and expected for the front-end development experience, I would argue.

However, I'm good with merging and maybe this can be added in a follow-up PR.

Thoughts?

wass3r (commented):
One more suggestion. Since we're now referencing `secrets.env`, can you rename `.env.example` to `secrets.env.example` or alternatively remove it. i personally like to keep it, but if we are going for consistency with other repos, i'm ok with removal.

DOCS.md
jbrockopp: Fixed 👍

DOCS.md
jbrockopp: Fixed 👍

wass3r (approved):

jbrockopp:
@wass3r @davidvader would you like me to update [the Start section](https://github.com/go-vela/ui/blob/docs/local/DOCS.md#start) of the new docs file to hot-reload content?

I'm thinking we could have both documented to offer them as solutions 👍 

Thinking I would more or less replicate [what was existing in the old docs](https://github.com/go-vela/ui/blob/master/.github/DOCS.md#developing).

wass3r:
Yes, that sounds great and sort of what I had in mind. Appreciate it!

davidvader:
That would be great!!! Thank you Jordan!

wass3r (approved):

KellyMerrick (approved):
lgtm

PR #227, opened 2020-10-28
feat(secret): add delete button

KellyMerrick:
* add 'Remove' button to secret page
* upon click 'Remove' changes to 'Confirm' with contrasting red/white color, and a 'Cancel' button appears
* upon cancel, reverts back to initial 'Remove' view
* upon 'Confirm', contrasting yellow 'Removing' button briefly displays (no screenshot) during api call and page redirects to secrets list page with success alert (\<name\> removed from \<type\> secrets.)

Example screenshots:
Default view:
image

Initial view of 'Confirm' and 'Cancel' after click on 'Remove'
image

Success alert with dynamic secret and type:
image


matt-fevold:
I like it. Would it make for a better user experience to swap the `confirm` and `cancel` buttons so the user can't accidentally double click it to remove it? Just throwing an idea out there :)

matt-fevold (commented):
I like it. Would it make for a better user experience to swap the `confirm` and `cancel` buttons so the user can't accidentally double click it to remove it? Just throwing an idea out there :)

davidvader:
Good point @matt-fevold ! What do you think @KellyMerrick ?

KellyMerrick:
I agree! Will do.

KellyMerrick:
Cancel and Confirm have been swapped.
image

matt-fevold (approved):
Nice! Also see you've got plenty of tests :)

davidvader (approved):
You could add some definition comments above the button functions, and update the comment above tryString, but other than that, Looks Good To Me! 🦖

PR #228, opened 2020-11-05
enhance(build): implemented additional build fields in recent build history

davidvader:
closes https://github.com/go-vela/community/issues/145

## Changes
- added host
- added branch
- added commit sha

Screen Shot 2020-11-05 at 3 57 23 PM


davidvader:
waiting on #229 then verifying style across browsers etc.

JordanSussman:
I do wonder if `host` should be updated to `agent` to avoid some confusion.

kneal:
I agree with @JordanSussman. I think _host_ makes sense to admins but the user would probably not understand what that field means.

davidvader:
I can agree with that, good point! I think worker is the most accurate alternative.

davidvader:
pr cleaned up, some changes from the demo have been reverted. rfr

jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #229, opened 2020-11-06
fix(build): fixing tooltip style

davidvader:
fix for pull #225 
simple style tweak

### DIFF
--- a/src/scss/_main.scss
+++ b/src/scss/_main.scss
@@ -703,6 +703,7 @@ nav {
   display: flex;
   flex-direction: row;
   justify-content: space-between;
+  width: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


src/scss/_main.scss
```diff
@@ -582,7 +582,7 @@ nav {
 }
 
 .build-history {
-  ul {
+  .previews {
```
davidvader, line 585: this was my mistake on suggesting a generic element rule, it ended up impacting the child `ul` inside the hover tooltip. 
Using a class instead 👍

jbrockopp (approved):
LGTM

KellyMerrick (approved):
lgtm

PR #230, opened 2020-11-11
chore: use secrets.env

wass3r:
with the changes introduced in #226 we now have to pass in `VELA_API` manually if you want to run local dev via `npm run dev`. This PR will override the standard location that `parcel` uses (`.env`) to use the newly added `secrets.env` file. So, as long as you define `VELA_API` in `secrets.env` as we did previously with `.env`, you will not need to pass it manually.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -112,8 +112,6 @@ This method of running the application enables hot-reloading of the code.
   #   image: ui:local
   #   networks:
   #     - vela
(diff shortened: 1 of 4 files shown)


davidvader (approved):
awesome 👍  big improvement to local dev with all of these changes

kneal (approved):
LGTM 🐬

PR #231, opened 2020-11-12
Adds autofocus to home page search

bmuenzenmeyer:
👋 

Addresses https://github.com/go-vela/community/issues/148

Adds the `autofocus` attribute to the home page search input, so a user can start typing immediately upon authenticating to Vela and find the repo they are looking for among their favorites.

vela_auto_focus

I tried adding a cypress test for this but two approaches both failed/timed out. 🤷 

```js
    it('home search bar should be focused', () => {
      cy.focused().then(($el) => $el.should('have.attr', 'data-test', 'home-search-input'))
    })
```

and

```js
    it('home search bar should be focused', () => {
      cy.get(":focus").should('have.attr', 'data-test', 'home-search-input')
    })
```

> I recognize this is unsolicited UX work so no hard feelings will be felt if this doesn't align with your standards.

`autofocus` is sometimes dicey because it will put the start of the keyboard focus into a different location, and you can only have one on each page. An alternative would be to implement a keyboard shortcut such as `/` to focus the search bar.


davidvader:
I like it! Let me double check with our accessibility guidelines to make sure this doesnt interfere with our tab-navigation goals but personally I would really like to see this! Thanks @bmuenzenmeyer ! 👍

davidvader:
Also when I get a free moment I'll take a look at those tests

davidvader:
@bmuenzenmeyer Circling back, we decided the feature doesnt interfere with our accessibility strategies, good idea 👍  What are your thoughts on adding autofocus to the global search bar for Source Repositories as well?

https://github.com/go-vela/ui/blob/eb2903896a0ec7222eeea67cb1036d2ef5013713/src/elm/Search.elm#L69-L73

bmuenzenmeyer:
I added it locally and it does not behave the same on the `account/source-repos` page - I think because of the async loading that occurs. 

If you navigate there, it does not focus.
If you refresh once there, it seems to work.

I suppose some Javascript could be added to stabilize the feature. Thoughts?

davidvader:
Noooo problem, thanks for trying that answers that. if I get some time I'll play around with it but this LGTM

davidvader (approved):

davidvader:
extended the functionality to the `/account/source-repos` page using Dom.focus
https://github.com/go-vela/ui/pull/235

wass3r (approved):

PR #232, opened 2020-11-13
fix(env): revert to use .env

wass3r:
unfortunately have to break convention with other repos and use `.env` for the UI project. the bundler (parcel) we're using is ignoring attempted overrides. @davidvader and I were trying to make it work. missed a couple of aspects during testing which made it appear to work :cone-of-shame:


davidvader (approved):

jbrockopp (approved):
LGTM

I think this is fine so long as we update the other repos to match 👍 

Maybe it's a blessing in disguise? I believe `.env` is the default so now we'll match that 😄

wass3r:
@jbrockopp i'll make a story for that - thanks

PR #233, opened 2020-11-16
feat(pipelines): view pipeline configurations and expanded templates

davidvader:
implements go-vela/community/issues/84


davidvader:
implemented, complete, just needs cypress tests 👍

davidvader:
Screen Shot 2020-11-16 at 3 11 43 PM
good to go! 👍

JordanSussman:
Is the note about "yaml fields will be sorted alphabetically when expanding templates" due to some relevant API endpoint returning the data in that format?

davidvader:
@JordanSussman Yeah when the compiler expands templates the marshal to json/yaml follows the go struct.
https://github.com/go-vela/types/blob/master/yaml/step.go

JordanSussman:
> 
> 
> @JordanSussman Yeah when the compiler expands templates the marshal to json/yaml follows the go struct. Which all of the structs are (for the most part) alphabetical.
> https://github.com/go-vela/types/blob/master/yaml/step.go

I don't think this should block the UI being implemented, but getting some issue created to look at removing that gotcha would be nice. A brief stackoverflow returned [this post](https://stackoverflow.com/questions/48293036/prevent-alphabetically-ordering-json-at-marshal) as a potential workaround within Go.

davidvader:
@JordanSussman I don't think it's blocking the UI implementation, the PR is RFR 👍 

but I agree I'd like to preserve order as well to remove the disclaimer in the UI. I am familiar with overloading the default marshaler so I brought a similar idea to the team last week in which we unmarshal to a map, derive the order, store it, remarshal to our struct and assign the order then use the ordering when writing it back to JSON which is very similar to the example you shared so thank you! It grew in complexity when using Gin's YAML marshaler so the team decided it wasn't a priority but I could revisit.

jbrockopp (approved):
LGTM

wass3r (approved):

PR #236, opened 2020-11-25
chore(refactor): move Msg into separate module, remove redundant function parameters

davidvader:
Proposal to refactor Msg a little bit by moving it into its own file and moving more code for Secrets and Build into their own components respectively.

### DIFF
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -66,13 +66,9 @@ import Msg exposing (Interval(..), Msg(..), RefreshData)
 import Nav
 import Pager
 import Pages exposing (Page(..))
-import Pages.Build.Logs
-    exposing
-        ( focusLogs
-        , getCurrentStep
-        )
+import Pages.Build.Logs exposing (focusLogs)
 import Pages.Build.Model
-import Pages.Build.Update exposing (expandActiveStep)
+import Pages.Build.Update exposing (expandActiveStep, getAllBuildSteps, getBuild, refreshBuild, refreshBuildSteps, refreshLogs)
 import Pages.Build.View
 import Pages.Builds exposing (view)
 import Pages.Home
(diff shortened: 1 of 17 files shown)


src/elm/Main.elm
```diff
@@ -1064,29 +847,29 @@ update msg model =
...
         BuildUpdate m ->
             let
-                ( newModel, action ) =
-                    Pages.Build.Update.update model m ( getBuildStepLogs, getBuildStepsLogs ) FocusResult
+                ( newModel, cmd ) =
+                    Pages.Build.Update.update model m
             in
             ( newModel
-            , action
+            , Cmd.map BuildUpdate cmd
```
davidvader, line 854: Using Cmd.map here allows us to dispatch Api calls from within Build.Update and then also receive the response in BuildUpdate

src/elm/Main.elm
```diff
@@ -719,72 +604,6 @@ update msg model =
                 Err error ->
                     ( { model | builds = { currentBuilds | builds = toFailure error } }, addError error )
 
-        StepResponse _ _ _ _ response ->
```
davidvader, line 722: moving the inbound API responses into Build.Update

src/elm/Main.elm
```diff
@@ -1319,9 +1068,9 @@ refreshPage model =
         Pages.Build org repo buildNumber focusFragment ->
             Cmd.batch
                 [ getBuilds model org repo Nothing Nothing Nothing
-                , refreshBuild model org repo buildNumber
-                , refreshBuildSteps model org repo buildNumber focusFragment
-                , refreshLogs model org repo buildNumber model.steps Nothing
+                , Cmd.map BuildUpdate <| refreshBuild model org repo buildNumber
```
davidvader, line 1071: we can still control page refreshes from Main.elm (or move it out someday) while using Cmd.map to map the responses back to the Build component

src/elm/Main.elm
```diff
@@ -1610,32 +1287,32 @@ viewContent model =
 
         Pages.AddOrgSecret engine _ ->
             ( "add " ++ engine ++ " org secret"
-            , Html.map (\m -> AddSecretUpdate engine m) <| lazy Pages.Secrets.View.addSecret model
```
davidvader, line 1613: renaming AddSecretUpdate to SecretsUpdate, more appropriately

src/elm/Msg.elm
```diff
@@ -0,0 +1,123 @@
...
+
+type alias RefreshData =
+    { org : Org
+    , repo : Repo
+    , build_number : Maybe BuildNumber
+    , steps : Maybe Steps
+    }
+
+
+type Msg
```
davidvader, line 61: by moving Msg into its own Module, we no longer have to pass generic Msgs down 15 function calls or bloat partial models with Msg aliases.
Now, we can import Msg(..) and use ToggleFavorite on ANY repo page

PR #253, opened 2020-12-31
feat(nav): moving Msg from Build/Pipeline back to Main.elm, prep work

davidvader:
prep work for go-vela/community#166

some changes here can be found in the open PR: https://github.com/go-vela/ui/pull/252

### What
this _reverts_ organizational changes that moved Msg from Main.elm to Build/Update.elm and reorganizes both Build and Pipeline page.
It also generalizes more steps code to prepare for the mass introduction of services and service logs.

### Why?
- Build Msgs must be moved back to Main.elm so that we can modularize logs for Steps/Services
- Pipeline Msgs must be moved back to Main.elm so that we can have BuildPipeline and Pipeline as two separate pages using the same msg bus.

PR #252, opened 2020-12-30
feat(nav): modularize logs, prep work

davidvader:
prep work for go-vela/community#166

no functional changes, but modularizes log helper functions using a new Resource type in Vela.elm
prepares for the introduction of services

PR #257, opened 2021-01-05
feat(nav): adding pipeline types and download yml button, prep work

davidvader:
prep work for go-vela/community#166

some changes here can be found in the open PRs: #252 #253

#### Changes
- adds a PipelineTemplates type for more easily tracking pipeline templates in the Model.
- adds a "Download vela.yml" button to Pipelines
Screen Shot 2021-01-05 at 1 37 41 PM

### DIFF
--- a/src/elm/Pages/Pipeline/View.elm
+++ b/src/elm/Pages/Pipeline/View.elm
@@ -44,8 +44,9 @@ import Vela
         ( Build
         , LogFocus
         , Org
-        , Pipeline
         , PipelineConfig
+        , PipelineModel
+        , PipelineTemplates
         , Repo
         , Step
         , Steps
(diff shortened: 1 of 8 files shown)


src/elm/Pages/Pipeline/View.elm
```diff
@@ -221,32 +222,55 @@ wrapPipelineConfigurationContent model cls content =
...
+    in
+    div [ class "actions" ] [ t, d ]
 
-        ( Loading, _ ) ->
-            text ""
 
-        _ ->
-            text ""
+velaYmlFileName =
+    "vela.yml"
```
kneal, line 268: A user could have a `.vela.yml` or `.vela.yaml` I don't think there is a clever way to know that. I personally think it's fine to just assume `.yml` or be opinionated and only download it to that file extension. 

But it might be an odd user experience to see `.yml` when the user stored their config in their repo as `.yaml`. Also, might not need to be fixed in this PR but might want to have a story to track fixing it.

kneal (approved):

src/elm/Pages/Pipeline/View.elm
davidvader: a story to track it, and adding that to the headers of the GET call might work.
but i TOTALLY agree, but as of now theres no way to tell which one theyre using.

jbrockopp (approved):
LGTM

PR #292, opened 2021-01-14
fix(css): adding margin to search bars

davidvader:
giving source repos and home page filter bar some breathing room

#### from
Screen Shot 2021-01-14 at 3 37 04 PM

#### to
Screen Shot 2021-01-14 at 3 37 15 PM

### DIFF
--- a/src/elm/Pages/Home.elm
+++ b/src/elm/Pages/Home.elm
@@ -72,7 +72,7 @@ view user filter { toggleFavorite, search } =
                 , a [ class "button", Routes.href Routes.SourceRepositories ] [ text "Source Repositories" ]
                 ]
     in
-    div [ Util.testAttribute "overview" ] <|
+    div [ class "overview", Util.testAttribute "overview" ] <|
         case user of
             Success u ->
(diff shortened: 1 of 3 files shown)


matt-fevold (dismissed):

wass3r (commented):
thanks - i've wanted to fix this as well. what do you think about going the other way? maybe adding top margin/padding to `.content-wrap` and removing top-margin/padding from various places where we are "fixing" that spacing?

davidvader:
> thanks - i've wanted to fix this as well. what do you think about going the other way? maybe adding top margin/padding to `.content-wrap` and removing top-margin/padding from various places where we are "fixing" that spacing?

sounds great, I'll do a run through and swap top spacing to `.content-wrap`

KellyMerrick (approved):
lgtm

wass3r:
@davidvader thanks for the edit.. i thought it was in more places :cone_of_shame:

but i think this will cause less duplication moving forward.

wass3r (approved):

PR #296, opened 2021-01-19
revert: purgecss upgrade from #293

wass3r:
purgecss v4 causes issues atm similar to the autoprefixer upgrade.  will need to address separately.


matt-fevold (approved):

KellyMerrick (approved):
lgtm

PR #304, opened 2021-01-26
feat: add visual alert for audit

wass3r:
closes https://github.com/go-vela/community/issues/191

image

Likely not the final form of this feature. Enough for MVP, probably.

One note: audit doesn't refresh unless it's the active tab, so you only see it on page load. It does disappear when a new build successfully starts (when you are on the builds tab) because it invalidates the condition by which the "alert" is added.

I lied, two notes: No tests (yet; hence draft status)

### DIFF
--- a/src/elm/Nav.elm
+++ b/src/elm/Nav.elm
@@ -76,6 +76,7 @@ type alias Tab =
     { name : String
     , currentPage : Page
     , toPage : Page
+    , isAlerting : Bool
     }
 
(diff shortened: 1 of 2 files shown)


wass3r:
I will add tests in a follow-up PR 😱

davidvader (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #314, opened 2021-02-02
fix(crumbs): change last segment to not be a link

wass3r:
Fixes issue primarily present in secrets where the last crumb is a link. The updated crumb design doesn't really accommodate that use. There's probably a couple of ways this could be fixed - this is one suggestion, adding "Add" as the last crumb when you are on a page to add a secret.

image
image

To consider.. changing "Add" to "Add Secret" or "Add Repo Secret"/"Add Org Secret".. that might eliminate the need for the heading on those pages, ie. "Add Repo Secret". Or, more invasively.. in the example's case, change "services" to "services Secrets" and keeping it as "Add", etc.

image

### DIFF
--- a/src/elm/Api.elm
+++ b/src/elm/Api.elm
@@ -353,7 +353,7 @@ patch api endpoint =
 -- ENTRYPOINT
 
 
-{-| try : default way to request information from and endpoint
+{-| try : default way to request information from an endpoint
 
     example usage:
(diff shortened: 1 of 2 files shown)


KellyMerrick (approved):
lgtm

jbrockopp (approved):
LGTM

PR #316, opened 2021-02-05
fix(proxy): switch to caddy

wass3r:
closes https://github.com/go-vela/community/issues/199

attempted various tweaks to nginx config to resolve, but wasn't able to fix the regression. there might be a solution with nginx, but i'm not sure what it is, atm. there's also likely potential to simplify backend redirects.

the switch to caddy does solve the issue but has some trade-offs (good and "bad") to be aware of:

- caddy alpine image is larger than the nginx one we used previously (~38MB vs ~14MB)
- caddy doesn't currently support brotli on-the-fly compression. bytes we transfer for our app is pretty reasonable. we will send an extra 40KB over the wire for largest asset (.js file) for example:
  ```
  134K Feb  5 11:05 static.77de5100.js.br
  175K Feb  5 11:04 static.77de5100.js.gz
  ```
  alternatively, we could explore pre-compressing assets down the road or wait until Caddy is able to implement a performant way to support brotli
- caddy's config and plugin system is arguably nicer
- caddy does native environment variable substitution
- easier potential to get rid of docker-entrypoint.sh script entirely (it's already been trimmed by half with this change) and other tweaks so that there is no file system modification (caddy does currently save a config file on startup, but there's a way to prevent it, afaict)
- built-in https handling might provide some nice options for testing w/ TLS locally


delta592 (approved):
looks good

delta592 (approved):
looks good

davidvader:
nice migration! 
pulled down and played around with it, stack looks good, local cli authflow working 👍  redirect working 👍  tests passing 👍 
LGTM!

davidvader (approved):

davidvader (approved):

jbrockopp (approved):
LGTM

PR #324, opened 2021-02-10
fix: reverting/fixing envsubst

davidvader:


kneal (dismissed):
LGTM 🐬

docker-entrypoint.sh
```diff
@@ -1,10 +1,16 @@
 #!/bin/sh
 set -eu
 
-for f in /srv/*.js
+for f in /usr/share/nginx/html/*.js
 do
     # shellcheck disable=SC2016
     envsubst '${VELA_API},${VELA_FEEDBACK_URL},{$VELA_DOCS_URL}' < "$f" > "$f".tmp && mv "$f".tmp "$f"
 done
+    envsubst '${VELA_API},${VELA_FEEDBACK_URL},{$VELA_DOCS_URL}' < "$f" > "$f".tmp && mv "$f".tmp "$f"
```
wass3r, line 9: The is there twice now unless it just looks like that on mobile.

davidvader: 👍

wass3r (approved):

PR #328, opened 2021-02-12
fix(repo): fixes error for non admin repo visitors

davidvader:
do not pre-load secrets unless visiting the tab explicitly

fixes: https://github.com/go-vela/community/issues/206

### DIFF
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -2665,7 +2665,7 @@ loadRepoSubPage model org repo toPage =
                             fetchSecrets o r
 
                         _ ->
-                            fetchSecrets org repo
+                            Cmd.none
                     ]
                 )
 


src/elm/Main.elm
```diff
@@ -2665,7 +2665,7 @@ loadRepoSubPage model org repo toPage =
                             fetchSecrets o r
 
                         _ ->
-                            fetchSecrets org repo
+                            Cmd.none
```
matt-fevold, line 2668: could this just be removed instead?

davidvader: > could this just be removed instead?

sadly I dont think so.
we use a `switch` to check if the page we are navigating to is the "Secrets" tab, and a `switch` in elm needs to either cover all possible values or have a default (the `_ ->` branch is the catch-all default).
we are just making the default do nothing (`Cmd.none`) so that we only `fetchSecrets` when the page we are headed to is the repo "Secrets" tab.

davidvader: it is basically the equivalent of
```
if toPage == Pages.RepoSecrets then
    fetchSecrets
else
    do nothing
```
we have each page split into its own switch so we can control the pre-load logic (IE, go to builds, and it pre-loads settings)

jbrockopp (approved):
LGTM

wass3r (approved):

PR #330, opened 2021-02-12
fix: stop UI duration timer when build is in a final state

davidvader:
fixes: https://github.com/go-vela/community/issues/208

Screen Shot 2021-02-12 at 1 43 24 PM

### DIFF
--- a/src/elm/Vela.elm
+++ b/src/elm/Vela.elm
@@ -98,6 +98,7 @@ module Vela exposing
     , encodeUpdateRepository
     , encodeUpdateSecret
     , encodeUpdateUser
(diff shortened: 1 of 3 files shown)


src/elm/Vela.elm
```diff
@@ -1131,8 +1137,11 @@ isComplete status =
         Error ->
             True
 
-        _ ->
-            False
+        Canceled ->
```
davidvader, line 1140: fixing an oversight by adding all types to this function

jbrockopp (approved):
LGTM

wass3r (approved):

PR #340, opened 2021-03-14
chore: dependency updates and more

wass3r:
This updates all but one dependency (`cypress`) and should close most of the open PRs from renovate. There will be a separate PR for Cypress. This PR also updates a couple of Elm dependencies not addressed yet by any of the open PRs.

One or more upgrades caused a snowball of changes also partially driven by a reported vulnerability via `npm audit` stemming from the now unmaintained `v1.x` version of Parcel.

Don't be scared of the LOC change :) - most of that is attributed to the package-lock file. I guess Parcel v2 gained some weight.

The following are some of the bigger changes:
* Bump Node to latest LTS version (`14.16.0`)
* Bump Parcel to use v2 (nightly) - this gets rid of one of the high security findings in `npm audit`

Other changes include:
* Move most dependencies to devDependencies
* Use `npm ci` in the Github Actions pipeline
* Remove NODE_ENV setting in CI - Parcel already injects that in prod builds
* Pin version of tools in Github Actions
* Change `.postcssrc.js` to JSON format (Parcel prefers it for caching)
* Remove unneeded dependencies from `package.json` as it's handled in Parcel
* Move type definition to same location as index.ts
* Fix lint issues


package.json
```diff
@@ -4,58 +4,56 @@
...
+    "postbuild:prod": "npm run copy:static",
+    "copy:static": "make-dir ./dist && ncp ./src/static/images ./dist/images",
+    "test": "elm-test",
+    "test:watch": "elm-test --watch",
     "test:cy": "start-test 8888 cypress",
     "test:cy-open": "start-test 8888 cypress:open",
     "cypress": "cypress run",
     "cypress:open": "cypress open",
-    "clean": "rimraf ./.cache ./dist",
+    "clean": "rimraf ./.parcel-cache ./.cache ./dist",
```
wass3r, line 56: One of the current, known issues with Elm + Parcel v2 - need to clean Parcel cache before a build.

src/static/index.ts
```diff
@@ -2,9 +2,11 @@
 //
 // Use of this source code is governed by the LICENSE file in this repository.
 
-import ClipboardJS from 'clipboard';
-import { App, Config, Elm, Flags, Theme } from '../elm/Main';
+// import types
+import { App, Config, Flags, Theme } from './index.d';
```
wass3r, line 6: Separated out types and moved type definition file in same folder. Maybe personal preference, but since this is the only TS occurrence in the project, I prefer the files close to each other.

package.json
```diff
@@ -4,58 +4,56 @@
...
+    "@parcel/transformer-elm": "2.0.0-nightly.2249",
+    "@parcel/transformer-sass": "2.0.0-nightly.627",
+    "axe-core": "4.1.3",
     "cypress": "5.6.0",
-    "cypress-axe": "0.12.0",
-    "elm-hot": "1.1.6",
+    "cypress-axe": "0.12.2",
+    "elm": "0.19.1-5",
+    "make-dir-cli": "2.0.0",
+    "ncp": "2.0.0",
```
wass3r, line 21: mainly added to deal with copying static image/favicon assets for builds. we used a parcel plugin previously and while the plugin has an update to work with Parcel v2, that plugin would be the only reason that would cause us to have a parcel configuration file at this time.

the alternative would be to use the `import .. from "url:./.."` syntax in Parcel v2 but then you have asset hashing to deal with, etc

jbrockopp (approved):
LGTM

davidvader (approved):

PR #369, opened 2021-04-09
chore: bump nginx to 1.19.1

wass3r:
Pinning to specific version of the nginx image and with that updating from Nginx `1.18.0` to `1.19.1` to be on latest.

This should mitigate: https://vuldb.com/?id.155282#:~:text=NGINX%20through%201.18.,since%2004%2F28%2F2020


jbrockopp (approved):
LGTM

KellyMerrick (approved):
lgtm

PR #389, opened 2021-04-28
fix: right align long tooltip fields

davidvader:
worker names are getting longer, right align tooltip fields

### before

Screen Shot 2021-04-28 at 2 26 13 PM

### after

Screen Shot 2021-04-28 at 2 25 22 PM


wass3r (approved):

jbrockopp (approved):
LGTM

PR #405, opened 2021-05-25
fix(form): fix overlap in non-stack scenario

wass3r:
fixes an issue with form label overflowing onto buttons making the buttons less accessible, kinda like..

```
+----form-control------+ +----<button>----+
| +-----------+ +------|-|----------------|---+
| |  <input>  | |        <label>          |   |
| +-----------+ +------|-|----------------|---+
+----------------------+ +----------------+
```
(not scaled to size)

it only allowed interactivity at the top and bottom of the button. 

the issue was observed in Firefox v89 on the Repo Settings page, specifically the Build Timeout setting.

### DIFF
--- a/src/scss/_forms.scss
+++ b/src/scss/_forms.scss
@@ -76,10 +76,14 @@
 
   &.-stack {
     flex-flow: row wrap;
+
+    .form-label {
+      flex: 1 0 100%;
+    }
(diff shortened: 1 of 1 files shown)


kneal (approved):
LGTM 🐬

PR #412, opened 2021-06-21
feat: add allow_command radio buttons to Secrets UI

JayCeeJr:
Related Issue: https://github.com/go-vela/community/issues/87

Added radio to select if secret should be allowed in commands.

Interestingly I found most of this code in another commit: https://github.com/go-vela/ui/commit/ea4e8056f40ee537c152add529d45d4ead22a931#diff-ebc97117a8ca6bd59c6e215aef88e99ca9daef03682fbab29858ece410e754c3R237-R256

The code from the commit above only worked for update, not for adding a secret. I was able to tweak it to work in both scenarios.

### DIFF
--- a/src/elm/Pages/Secrets/Form.elm
+++ b/src/elm/Pages/Secrets/Form.elm
@@ -244,18 +244,18 @@ radio value field title msg =
 -}
 viewAllowCommandCheckbox : SecretForm -> Html Msg
 viewAllowCommandCheckbox secretUpdate =
-    section [ class "type", Util.testAttribute "" ]
-        [ h4 [ class "field-header" ]
-            [ text "Allow Commands"
+    section [ Util.testAttribute "" ]
+        [ div [ class "form-control" ]
+            [ strong [] [ text "Allow Commands"
             , span [ class "field-description" ]
                 [ text "( "
                 , em [] [ text "\"No\" will disable this secret in " ]
                 , code [] [ text "commands" ]
                 , text " )"
                 ]
-            ]
+            ]]
         , div
-            [ class "form-controls", class "-row" ]
+            [ class "form-controls", class "-stack" ]
             [ radio (Util.boolToYesNo secretUpdate.allowCommand) "yes" "Yes" <| OnChangeAllowCommand "yes"
             , radio (Util.boolToYesNo secretUpdate.allowCommand) "no" "No" <| OnChangeAllowCommand "no"
             ]
--- a/src/elm/Pages/Secrets/Form.elm
+++ b/src/elm/Pages/Secrets/Form.elm
@@ -5,7 +5,8 @@ Use of this source code is governed by the LICENSE file in this repository.
 
 
 module Pages.Secrets.Form exposing
-    ( viewEventsSelect
+    ( viewAllowCommandCheckbox
+    , viewEventsSelect
     , viewHelp
     , viewImagesInput
     , viewNameInput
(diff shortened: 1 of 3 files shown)


JayCeeJr:
I looked into writing Cypress tests, however there were no tests around adding or updating secrets. Only retrieving and deleting. 
I'm not very familiar with Cypress, and not sure if it is possible to assert XHR payloads, since this change only flips a boolean in the payload.

src/elm/Pages/Secrets/Form.elm
```diff
@@ -219,6 +224,44 @@ viewImagesInput secret imageInput =
...
+            []
+        , label [ class "form-label", for <| "radio-" ++ field ] [ strong [] [ text title ] ]
+        ]
+
+
+{-| allowCommandCheckbox : renders checkbox inputs for selecting allowcommand
+-}
+viewAllowCommandCheckbox : SecretForm -> Html Msg
+viewAllowCommandCheckbox secretUpdate =
+    section [ class "type", Util.testAttribute "" ]
```
wass3r, line 247: I don't believe there is a `.type` CSS class. We can probably remove it.

src/elm/Pages/Secrets/Form.elm
```diff
@@ -219,6 +224,44 @@ viewImagesInput secret imageInput =
...
+        [ h4 [ class "field-header" ]
+            [ text "Allow Commands"
+            , span [ class "field-description" ]
+                [ text "( "
+                , em [] [ text "\"No\" will disable this secret in " ]
+                , code [] [ text "commands" ]
+                , text " )"
+                ]
+            ]
```
wass3r, line 256: I think the `h4` is part of the reason for the header being nudged over in this case. Should be able to make it more consistent by modelling it after: https://github.com/go-vela/ui/blob/067939009c9c405ff333a4d9879418d612ce8cca/src/elm/Pages/Secrets/Form.elm#L147-L155

src/elm/Pages/Secrets/Form.elm
```diff
@@ -219,6 +224,44 @@ viewImagesInput secret imageInput =
...
+        , div
+            [ class "form-controls", class "-row" ]
```
wass3r, line 258: I understand the appeal of using `-row` here. However, I don't feel we adequately account for this from a styling perspective yet. The spacing between the radio selection controls and the labels are not distributed well to create sensible groupings in a `-row` scenario.

I would propose using `-stack` for now, which also keeps it consistent with the radio control layout on the repo settings page for controlling visibility (i know it's not a Yes/No scenario and we have a bit more content for the options).

wass3r (changes requested):
Thanks for the PR! Tested out locally and works great.

I did notice an issue with the "heading" being nudged in (see [1] in the screenshot) in FF and Chrome and added some comments for a potential fix inline with the code.

Also added some suggestions for [2] in the screenshot inline with the respective code. Something else to consider here is whether we should consider turning that into a checkbox instead of a distinct Yes/No selection. Curious on your thoughts.

secret_feedback

JayCeeJr:
I fixed the alignment issue. 

As for using a checkbox over the Yes / No radio buttons, I think the radio buttons look better.
If there was only one checkbox, I think we would either remove the header, or have redundant descriptions in the header and text next to the checkbox.

wass3r (approved):
TY 👍🏼

jbrockopp (approved):
LGTM

PR #425, opened 2021-07-30
feat: add actions menu to build list

JayCeeJr:
Add restart build button to build list 

Related Issue https://github.com/go-vela/community/issues/143
Screen Shot 2021-08-02 at 1 57 32 PM
Screen Shot 2021-08-02 at 1 27 11 PM


JayCeeJr:
Since I was in this code, I added the name of the tag and the pull request number.

Screen Shot 2021-08-04 at 9 07 55 AM

wass3r (changes requested):
thanks for the PR!

one initial ask: i would prefer if the addition for showing the tag (though appreciated) was a separate PR to keep the PR focused.

JayCeeJr:
@wass3r  I reverted the tag and PR commit and opened a new PR

wass3r:
one of the other PRs must have introduced a conflict 😿

wass3r:
apologies for the delay. any chance we can address this visual discrepancy? (it's also visible in your screenshots)

actions

JayCeeJr:
> apologies for the delay. any chance we can address this visual discrepancy? (it's also visible in your screenshots)
> 
> actions

I'm not sure how to accomplish keeping the space consistent, the menu is absolute positioned, and the rest is flex. The error adds an additional row which shifts the content up.

One solution would be to change the info box to `.build .info {justify-content: flex-start;}` (main.scss:517) however, i think that looks worse.
Screen Shot 2021-09-20 at 9 25 29 AM


Another would be to add `.build .info {min-height: 2em;}` (main.scss:532), which also doesn't look great.
Screen Shot 2021-09-20 at 9 31 13 AM

wass3r:
what if we throw `position: relative` on the first row and move `details` in it? probably have to adjust a few things, like position on `details` element and such. haven't tried too much, but seems feasible from afar?

i can also take a crack at it after merge.

src/scss/style.scss
```diff
@@ -1,4 +1,4 @@
-@charset 'utf-8';
+@charset "utf-8";
```
davidvader, line 1: was this change intentional?

JayCeeJr: Must have been the IDE trying to be helpful

wass3r (approved):
sorry for the delay but thanks for the work (again)!

tbh, not a huge fan of it being off-centered in the 2-row version, but it is consistent now 👍🏼

jbrockopp (approved):
LGTM

PR #449, opened 2021-09-23
chore(ci): use latest ubuntu to fix GH actions

wass3r:
Seems that using `ubuntu-16.04` is not supported anymore. It was pinned previously due to issues with Cypress. Attempting to use `ubuntu-latest` here to see if we can get away with it :)


jbrockopp (approved):
LGTM

PR #453, opened 2021-10-01
enhance(logs): handle larger logs

davidvader:
solution for https://github.com/go-vela/community/issues/261
and other issues with trying to view MASSIVE logs in the UI.

This PR accomplishes:
- imposing a limit to the log size rendered in the UI
- removes the dependency the download button has on the log being decoded already
- moves the "step logs not written to" message to inside the log window, so a user can start following a step before it has produced any logs

## log size limit
this PR imposes a limit to the size of logs the UI will attempt to base64 decode and render. 

* the limit as it stands in the PR is set to `5mb` as represented by 1000 * 1000 bytes.
* The limit is currently _hardcoded_ but this could certainly be moved to `Flags` and be configured on runtime.
* * The limit could also be a configuration set on the backend by the admins, however this seemed like a good place to start.

#### after:
Screen Shot 2021-10-01 at 4 15 10 PM

## remove download button dependency

This enhancement also decouples the base64 bottleneck from the download button, so users can easily download large logs using the "download step logs" button. Before, the button would not render until the base64 decode was complete, which in some cases would never finish.

## empty logs can now be followed

this is just a side effect to cleaning up the code and having a concept of a "default log message" when logs are empty or too big.

#### before (no play button):
Screen Shot 2021-10-01 at 4 17 58 PM

#### after (play button):
Screen Shot 2021-10-01 at 4 15 28 PM


src/elm/Vela.elm
```diff
@@ -1549,22 +1550,34 @@ type alias Log =
...
+                id
+                step_id
+                service_id
+                build_id
+                repository_id
+                data
+                -- "decodedLogs"
+                ""
+                -- "size"
+                (Bytes.Encode.getStringWidth data)
```
davidvader, line 1573: I had to modify the log decoder syntax so that we could use the value for `data` to supply the value for `size`.

src/elm/Util.elm
```diff
@@ -499,3 +502,36 @@ getNameFromRef s =
...
+    1000 * kilobyte
+
+
+{-| formatFilesize : returns a file size in bytes as a human readable string.
+Defined as a helper function to make it easier to configure bases, units etc.
+see: <https://package.elm-lang.org/packages/basti1302/elm-human-readable-filesize/latest/Filesize>
+-}
+formatFilesize : Int -> String
+formatFilesize =
+    Filesize.format
```
davidvader, line 549: this helper seems unnecessary but might come in handy if we want to customize the file formatter

src/elm/Pages/Build/View.elm
```diff
@@ -796,9 +774,9 @@ expandAllButton expandAll org repo buildNumber =
 {-| logsHeader : takes number, filename and decoded log and renders logs header
 -}
 logsHeader : LogsMsgs msg -> String -> String -> String -> String -> Html msg
-logsHeader msgs resource number fileName decodedLog =
+logsHeader msgs resource number fileName logData =
     div [ class "logs-header", class "buttons", Util.testAttribute <| "logs-header-actions-" ++ number ]
-        [ downloadLogsButton msgs.download resource number fileName decodedLog ]
+        [ downloadLogsButton msgs.download resource number fileName logData ]
```
davidvader, line 779: changed this to use the raw data, offsetting the base64 decode work to the actual button click.

src/elm/Pages/Build/View.elm
```diff
@@ -655,32 +645,20 @@ viewLines : FocusLine msg -> Resource -> ResourceID -> LogFocus -> String -> Boo
 viewLines focusLine resource resourceID logFocus decodedLog shiftDown =
     let
         lines =
-            if not <| logEmpty decodedLog then
```
davidvader, line 658: we no longer need this code because a decodedLog will never be empty.
if the log were empty, it would be decoded with the helpful message and be safe to render.

src/elm/Pages/Build/View.elm
```diff
@@ -615,32 +613,24 @@ viewServiceLogs msgs shift rm service =
...
-            downloadFileName org repo buildNumber resource resourceID
-    in
     div
         [ class "logs"
         , Util.testAttribute <| "logs-" ++ resourceID
         ]
     <|
         case Maybe.withDefault RemoteData.NotAsked maybeLog of
-            RemoteData.Success _ ->
-                if logEmpty decodedLog then
```
davidvader, line 632: we no longer need this code because a decodedLog will never be empty.
if the log were empty, it would be decoded with the helpful message and be safe to render.

src/elm/Pages/Build/Logs.elm
```diff
@@ -294,6 +333,27 @@ downloadFileName org repo buildNumber resourceType resourceNumber =
     String.join "-" [ org, repo, buildNumber, resourceType, resourceNumber ]
 
 
+{-| logSizeLimit : returns the maximum size for log data, in bytes.
+-}
+logSizeLimit : Int
+logSizeLimit =
+    Util.megabyte * 5
```
davidvader, line 340: here we define the actual size limit.

should this be a runtime flag, or should this just be imposed by the code.

src/elm/Pages/Build/Logs.elm
```diff
@@ -158,6 +158,69 @@ getLog resource get logs =
...
+
+
+{-| logSizeExceededMessage : returns the default message when a log exceeds the size limit.
+-}
+logSizeExceededMessage : String
+logSizeExceededMessage =
+    "The data for this log exceeds the size limit ("
+        ++ Util.formatFilesize logSizeLimit
+        ++ ").\n"
+        ++ "To view this log use the CLI or click the 'download' link in the top right corner of this step (downloading may take a few moments, depending on the size of the file)."
```
davidvader, line 221: this message could definitely be tweaked. 
what would provide the most value to the user?

src/elm/Pages/Build/Logs.elm
```diff
@@ -158,6 +158,69 @@ getLog resource get logs =
...
+    { log
+        | decodedLogs =
+            if isEmpty log then
+                logEmptyMessage
+
+            else if exceedsSizeLimit log then
+                logSizeExceededMessage
+
+            else
+                Util.base64Decode log.rawData
```
davidvader, line 203: moving the `Util.base64Decode` call behind the size limit has improved the UI performance drastically. 

base64 decoding accounts for the majority of page-unresponsiveness when viewing logs

src/elm/Main.elm
```diff
@@ -374,7 +375,7 @@ type Msg
     | ShowHideHelp (Maybe Bool)
     | ShowHideIdentity (Maybe Bool)
     | Copy String
-    | DownloadFile String String String
+    | DownloadFile String (String -> String) String String
```
davidvader, line 378: adding a string transformation function so the download Msg can be supplied an expensive function that is only called when the user executes the code (clicks the button)

PR #489, opened 2021-12-07
fix(builds): collapse actions dropdown when clicking anywhere else

davidvader:
closes https://github.com/go-vela/community/issues/452

## BEFORE

https://user-images.githubusercontent.com/48764154/145068956-c2ccdb21-f315-4c00-8af1-c138536109ea.mov

## AFTER

https://user-images.githubusercontent.com/48764154/145068973-26923764-5033-4484-b8ca-7da05b22eb8a.mov


src/elm/Pages/Build/View.elm
```diff
@@ -271,7 +271,7 @@ viewPreview msgs openMenu showMenu now zone org repo build =
         message =
             [ text <| "- " ++ build.message ]
 
-        id =
+        buildId =
```
davidvader, line 274: removing bad var usage of imported Html.attributes function `id`

jbrockopp (approved):
LGTM

src/elm/Pages/Build/Model.elm
```diff
@@ -58,7 +58,7 @@ type alias Msgs msg =
     , logsMsgs : LogsMsgs msg
     , restartBuild : RestartBuild msg
     , cancelBuild : CancelBuild msg
-    , toggle : Int -> Maybe Bool -> msg
+    , toggle : Maybe Int -> Maybe Bool -> msg
```
wass3r, line 61: what's the reason for turning it into a Maybe?

davidvader: the onMouseDown function expects a Maybe Bool

https://github.com/go-vela/ui/blob/b339ad66dace720f77923a25f150bc1fda25301b/src/elm/Main.elm#L2256

davidvader: reason being the click could be from somewhere else, not the id button

wass3r: but we don't have an instance where we don't pass an id, right? i could have missed it

davidvader: `onMouseDown` accepts a `Maybe Bool` by design because all of our toggles are embedded to click with `Nothing` to indicate a "toggle `VALUE` to `!VALUE`". 

`onMouseDown` takes any click registered _outside_ of the target ID and dispatches a `Just False` triggerToggleMsg.
 
Basically, its a way to use a generic TriggerMsg that uses `Nothing` to indicate a bit flip, and `Just X` to indicate a manual close/open of a tooltip/dropdown.

wass3r: `Maybe Bool` makes sense. Was confused on the `Maybe Int` since we only use `Just X` and never `Nothing` (that i could tell).

wass3r (approved):

src/elm/Pages/Build/Model.elm
davidvader: ohhhh! i completely misread your initial question, you asked about maybe int and not maybe bool. sorry about that.

I changed `Int` to a `Maybe Int` because normally a build ID is passed in every time:
https://github.com/go-vela/ui/blob/master/src/elm/Pages/Build/View.elm#L194

but when we use this Msg function in `onMouseDown`, we don't have a build ID, so we need a `Maybe Int` to be able to pass in `Nothing` and consider that the "close all":
https://github.com/go-vela/ui/pull/489/files#diff-e0e84d060b84a8fe3dc523e02e0fe5a5ab03d46ea3895eba481f6750d3089de9R682-R689

there are 2 `unwrap` calls, the first one you see is unwrapping `Nothing` into `[]` which represents `onMouseDown` being clicked outside of the build actions menu and receiving `Nothing`, the `Just` unwrap indicates it was actually clicked and executes normally

PR #493, opened 2021-12-22
fix: actions menu and build animation overflow bug

davidvader:
adding a width/height wrapper in conjunction with `overflow-x: clip` to cause only the animation particles to get hidden when leaving the build-container

also, `z-index: 9999` on the actions menu to place it in front of the svg "particles"

### before

Screen Shot 2021-12-07 at 11 03 35 AM

### after

Screen Shot 2021-12-22 at 1 01 56 PM


wass3r (approved):

jbrockopp (approved):
LGTM

PR #494, opened 2021-12-22
feat: add build limit to repo settings page

jbrockopp:
Dependent on https://github.com/go-vela/server/pull/558 and https://github.com/go-vela/types/pull/219

Part of the effort for https://github.com/go-vela/community/issues/27 and https://github.com/go-vela/community/issues/162

Based off of https://github.com/go-vela/community/blob/master/proposals/2020/12-16_rate-limiting.md

This adds support for the `BuildLimit` field that can be set for a repo in the [go-vela/server](https://github.com/go-vela/server):

image

When a user attempts to modify this value, a message is displayed:

image

If a user tries to provide a value outside what we consider valid (`1` - `30`), then the `update` button grays out:

image

Additionally, a new runtime `VELA_MAX_BUILD_LIMIT` environment variable has been added to the UI.

This will be used to control the behavior when attempting to update the build limit for a repo on the settings page.

By default, it comes with a value of `30` like the [go-vela/server](https://github.com/go-vela/server) does.

The above functionality still works if the limit is updated to a different value (example with `50` as limit):

image


cypress/integration/repo_settings.spec.js
```diff
@@ -75,6 +79,33 @@ context('Repo Settings', () => {
...
+      cy.get('@repoLimitInput').should('not.have.value', 'cat');
+      cy.get('@repoLimitInput').type('{selectall}12cat34');
+      cy.get('@repoLimitInput').should('have.value', '1234');
+    });
+
+    it('clicking update on build limit should update limit and hide button', () => {
+      cy.get('[data-test=repo-limit]').as('repoLimit');
+      cy.get('[data-test=repo-limit] input').as('repoLimitInput');
+      cy.get('@repoLimitInput').should('be.visible').clear();
+      cy.get('@repoLimitInput').type('{selectall}80');
```
davidvader, line 102: `{selectall}` is cool

docker-entrypoint.sh
```diff
@@ -4,7 +4,7 @@ set -eu
 for f in /usr/share/nginx/html/*.js
 do
     # shellcheck disable=SC2016
-    envsubst '${VELA_API},${VELA_FEEDBACK_URL},{$VELA_DOCS_URL}' < "$f" > "$f".tmp && mv "$f".tmp "$f"
+    envsubst '${VELA_API},{$VELA_DOCS_URL},${VELA_FEEDBACK_URL},${VELA_MAX_BUILD_LIMIT}' < "$f" > "$f".tmp && mv "$f".tmp "$f"
```
davidvader, line 7: is this a typo?
```suggestion
    envsubst '${VELA_API},${VELA_DOCS_URL},${VELA_FEEDBACK_URL},${VELA_MAX_BUILD_LIMIT}' < "$f" > "$f".tmp && mv "$f".tmp "$f"
```

jbrockopp: fixed 👍

src/elm/Main.elm
```diff
@@ -497,6 +503,18 @@ update msg model =
...
 
+        ChangeRepoLimit limit ->
+            let
+                newLimit =
+                    case String.toInt limit of
+                        Just t ->
+                            Just t
+
+                        Nothing ->
+                            Just 0
```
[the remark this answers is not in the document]
jbrockopp: this should be addressed 👍 

I used `Maybe.withDefault` like you asked but had to make an additional change due to that.

plyr4 (approved):
LGTM - api errors end up setting the build limit to 0, which was confusing from a user experience perspective but looks good from the UI

dtanner (approved):

PR #502, opened 2022-01-19
fix(favorites): new favorite logic for enable repo

ecrupper:
[Related Issue](https://github.com/go-vela/community/issues/173)

With ToggleFavorites, repos that were previously favorited would become unfavorited upon enabling the repo. This is most likely not the desired behavior, so I added the ability to add a repo to a user's favorites or maintain its status as a favorite when a user enables it.


src/elm/Favorites.elm
```diff
@@ -93,6 +93,25 @@ updateFavorites user favorite =
             ( [], False )
 
 
+{-| enableFavorite : takes current user and favorite key and adds favorite to list of favorites
+-}
+enableFavorite : WebData CurrentUser -> String -> ( List String, Bool )
+enableFavorite user favorite =
```
plyr4, line 99: thoughts on making this `addFavorite`?

```suggestion
addFavorite : WebData CurrentUser -> String -> ( List String, Bool )
addFavorite user favorite =
```

ecrupper: yeah for sure — I wasn't sure about the naming norms regarding `AddFavorite` and `addFavorite` since there was `ToggleFavorite` and `updateFavorite`, but I think `addFavorite` definitely makes more sense

GregoryDosh (dismissed):
I don't have a super firm graps on the inner workings of ELM like I do React, but this looks pretty reasonable and looks like has already gone through a few revisions. Going to give this a 👍  and others can chime in if there is something else I'm missing.

wass3r (changes requested):
gave this a whirl and it looks like it keeps the repo favorited as described. however, it still triggers a toast message saying that it unfavorited the repo when it didn't:

image

jbrockopp (approved):
LGTM

PR #505, opened 2022-01-31
fix: a couple of UI fixes

wass3r:
The first fix is an issue introduced in v0.13.0-rc. The timestamp addition grew the filter bar vertically and started really hugging the list controls. Fix was to move margins from previous sole filter component to wrapper container that was added.

before:
build_bar_before
after:
build_bar_after

The second fix is for the recent build widget that has been there for a few versions. The text was not aligned on the left side - neither was the first item (if it was active). Also reduced distance between items slightly to match line weight used throughout.

before:
recent_builds_before
after:
recent_builds_after

Also removing some unused imports.

### DIFF
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -66,7 +66,7 @@ import Html.Attributes
         , type_
         )
 import Html.Events exposing (onClick)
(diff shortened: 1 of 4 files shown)


jbrockopp (approved):
LGTM

kaymckay (approved):
lgtm

PR #517, opened 2022-02-16
chore(node): bump node and simplify ci

wass3r:
bumps node version to keep in line with https://github.com/go-vela/ui/pull/511

also simplifies Github Actions flow to pull node version from nvmrc in root of project instead of defining separately in each action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,9 @@ name: ci
 # trigger on pull_request or push events
 on:
   push:
-    branches:
   pull_request:
 
 env:
-  NODE_VERSION: '16.13.2'
   CI: 1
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -8,11 +8,9 @@ name: ci
 # trigger on pull_request or push events
 on:
   push:
-    branches:
```
wass3r, line 11: yaml validation didn't like this and this change would make this action consistent with what we do elsewhere, eg. https://github.com/go-vela/server/blob/ea8fadd3c76883df252fb3c3b60681deebd85e79/.github/workflows/test.yml#L4-L7

jbrockopp (approved):
LGTM

PR #526, opened 2022-03-07
chore: remove changelog

kaymckay:


ecrupper (dismissed):

wass3r (changes requested):
so, there is some inconsistency here between the other repos. in this case `release.yml` actually doesn't generate the release notes. in this case it's `publish.yml`.

kaymckay:
@wass3r can we rename it to publish.yml then?

wass3r:
works for me

wass3r (approved):

plyr4 (approved):

PR #535, opened 2022-03-24
Feat: Add ability to copy secret to clipboard

kaymckay:
Part 2 of https://github.com/go-vela/community/issues/516

Copies the secret how it should be formatted for yml

For example: 
copy button on octocat/hello-world/demo_secret would copy
```
  - name: demo_secret
    key:  octocat/hello-world/demo_secret
    engine: native
    type: repo
```

Also adds tests for this and the key field added to the table

### DIFF
--- a/src/elm/Pages/Secrets/View.elm
+++ b/src/elm/Pages/Secrets/View.elm
@@ -29,6 +29,7 @@ import Vela
         ( Secret
         , SecretType(..)
         , Secrets
+        , secretToKey
         , secretTypeToString
         , secretsErrorLabel
         )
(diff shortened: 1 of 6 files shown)


plyr4 (approved):
lgtm - nice

src/elm/Pages/Secrets/View.elm
```diff
@@ -429,6 +451,44 @@ updateSecretHref type_ secret =
...
+            yaml ++ "shared"
+
+
+{-| copyButton : copy button that copys secret yaml to clipboard
+-}
+copyButton : String -> Html Msg
+copyButton copyYaml =
+    button
+        [ class "copy-button"
+        , attribute "aria-label" <| "copy secret yaml '" ++ copyYaml ++ "' to clipboard "
```
wass3r, line 479: i think we should simplify the `aria-label` text. i think when screen readers read this aloud it will not be super helpful after "copy secret yaml". maybe just leave out the `copyYaml` part?

wass3r (changes requested):
one minor suggestion, but looks and works good otherwise! thank you!

wass3r (approved):

wass3r (approved):

jbrockopp (approved):
LGTM

PR #552, opened 2022-05-20
fix(nginx): strip version number from server header

jbrockopp:
This change strips the version number from the `Server` response header:

http://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens

This was flagged, as a low finding, by Cargill's security team during a penetration test of the system.

Their argument is that knowing the version information enables someone to exploit the product.

If the version is known, someone could attempt to use a known vulnerability, for that version, against the system.

## Before

image

## After

image


cognifloyd (approved):

wass3r (approved):

JohnTrammellTarget (commented):
lgtm

PR #575, opened 2022-08-19
fix: pipeline download link

plyr4:
fixes the download link for the revised pipeline tab.
the button was pulling from decodedData that was never being updated in the model.


wass3r (approved):

jbrockopp (approved):
LGTM

PR #640, opened 2023-03-31
fix(deployments): point commit link to clone/tree/ref instead of api deployment

plyr4:
see title

closes https://github.com/go-vela/community/issues/598


ecrupper (approved):

wass3rw3rk (approved):
wonder if we should do the build event replace of `_` during decode to avoid replacing everywhere?

PR #642, opened 2023-03-31
fix(build): fetch resource updates when they are incomplete

plyr4:
see title

fixes a bug where a build completes, the UI gets the "completed" build status and then stops refreshing the build steps/services to save on resources. 
this just changes the code to check if the build OR any steps/services are "incomplete"

closes https://github.com/go-vela/community/issues/777


wass3rw3rk (approved):
lgtm

PR #643, opened 2023-03-31
fix(deployments): fetch deployment when tab switching to PromoteDeployment page

plyr4:
see title

fixes a bug where clicking "Redeploy" or getting redirected to the Promote Deployment page (redeploy) would not fetch the deployment nor would it pre-populate with the deployment we already have available.

closes https://github.com/go-vela/community/issues/725

### DIFF
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -3135,9 +3135,6 @@ loadRepoSubPage model org repo toPage =
         fetchSecrets o r =
             Cmd.batch [ getAllRepoSecrets model "native" o r, getAllOrgSecrets model "native" o ]
 
-        _ =
-            Debug.log "toPage" toPage
-
         -- update model and dispatch cmds depending on initialization state and destination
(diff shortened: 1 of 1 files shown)


src/elm/Main.elm
```diff
@@ -3135,6 +3135,9 @@ loadRepoSubPage model org repo toPage =
         fetchSecrets o r =
             Cmd.batch [ getAllRepoSecrets model "native" o r, getAllOrgSecrets model "native" o ]
 
+        _ =
+            Debug.log "toPage" toPage
```
wass3rw3rk, line 3139: i don't think we want to keep this

wass3rw3rk (dismissed):

src/elm/Main.elm
plyr4: ooo good catch, fixed. thank you!

KellyMerrick (approved):
lgtm

wass3r (approved):

PR #656, opened 2023-04-26
feat: adding /pulls shortcut route for viewing pull_request builds

plyr4:
see title, this adds a route shortcut `/org/repo/builds/pulls` for easily viewing only pull request builds.
its basically a shortcut to using `?event=pull_request`  
the idea came from the ease of moving between the SCM and a hosted vela instance. 

if this gets approvals, i will most likely add `/tags` and `/deployments` 

Screenshot 2023-04-26 at 10 33 50 AM

_**Note:**_ it is adding a "reserved route"

PR #663, opened 2023-05-16
feat: adding /tags shortcut route for viewing tag builds

plyr4:
see: https://github.com/go-vela/ui/pull/656
same change, but for tag event
this adds a route shortcut `/org/repo/builds/tags` for easily viewing only tag builds.
its basically a shortcut to using `?event=tag`

i find myself swapping between github and vela, and like the convenience of being able to swap the host and like when the vela url matches with filtered content.

Screenshot 2023-05-16 at 9 33 15 AM

_**Note:**_ it is adding a "reserved route"

### DIFF
--- a/cypress/integration/builds.spec.js
+++ b/cypress/integration/builds.spec.js
@@ -348,4 +348,38 @@ context('Builds', () => {
       cy.url().should('contain', '?event=comment');
     });
   });
+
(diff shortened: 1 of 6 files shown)


KellyMerrick (approved):
lgtm

timhuynh94 (approved):

PR #666, opened 2023-06-02
refactor: use ResourceType in place of Resource

plyr4:
`Resource` already exists as a struct, this makes it more clear that its the "ResourceType" and refers to a string.

PR #667, opened 2023-06-02
feat: parse and render links in logs

plyr4:
closes https://github.com/go-vela/community/issues/818
~a lot of these changes are refactors from https://github.com/go-vela/ui/pull/666, which should get merged first.~ moved 

this adds link parsing to log rendering :) 

it also cleans up a couple function names, and adds a `LogLine` struct that is meant to make log rendering more flexible. i used a similar approach to create a proof of concept for https://github.com/go-vela/community/issues/243, so this will play a key role in extending the logs even more.

### Before
Screenshot 2023-06-02 at 3 44 07 PM

### After
Screenshot 2023-06-02 at 3 44 31 PM


src/scss/_main.scss
```diff
@@ -979,15 +979,6 @@ details.build-toggle {
   background: var(--color-focus);
 }
 
-.line a {
```
plyr4, line 982: this is not used, and was also conflicting with the new <a> tag. it should be safe to remove.

plyr4: i believe it was used prior to changing the line numbers to `<button>`

src/elm/Pages/Build/Logs.elm
```diff
@@ -385,9 +456,164 @@ defaultPosition =
...
+    Html.span (styleAttributes chunk.style) <|
+        parseLinksInChunk chunk
+
+
+{-| styleAttributes : takes Ansi.Log.Style and renders it into ANSI style Html attributes
+see: <https://package.elm-lang.org/packages/vito/elm-ansi>
+this function has been pulled in unmodified because elm-ansi does not expose it
+-}
+styleAttributes : Ansi.Log.Style -> List (Html.Attribute x)
+styleAttributes style =
```
plyr4, line 491: i'd rather not pull these in as static code, but the ansi module is too restricting and converts to html without much room for customization. 

however, the module's parsing behind the scenes is incredible, so i dont mind pulling these in.

src/elm/Pages/Build/Logs.elm
```diff
@@ -342,27 +349,91 @@ downloadFileName org repo buildNumber resourceType resourceNumber =
...
+        )
+        split
+
+
+{-| linkEscapeCharacters : list of string escape characters that delimit links.
+for example "<https://github.com"> should be split from the quotes, even though " is a valid URL character (see: see: <https://www.rfc-editor.org/rfc/rfc3986#section-2>)
+-}
+linkEscapeCharacters : List String
+linkEscapeCharacters =
+    [ "'", " ", "\"", "\t", "\n" ]
```
plyr4, line 413: there might be more values to add here, i nearly forgot `\t` and caught it during testing.

src/elm/Pages/Build/Logs.elm
```diff
@@ -342,27 +349,91 @@ downloadFileName org repo buildNumber resourceType resourceNumber =
...
+                    logView =
+                        -- we need to ignore extra empty lines generated by Ansi.Log.update
+                        case List.head <| Array.toList ansiLogModel.lines of
+                            Just logLine ->
+                                -- we may need to extend this to accept user preferences
+                                -- for example, a user may choose to not render links
+                                processAndViewLine logLine
+
+                            Nothing ->
+                                text ""
```
plyr4, line 376: i expect this section of code to get refactored when we introduce of log timestamps in https://github.com/go-vela/community/issues/243

this is where we would parse metadata and check the user model then do custom rendering.

plyr4:
RFR, moved some code around so rendering is only in View.elm. it should be easier to review once https://github.com/go-vela/ui/pull/666 is reviewed/merged.

plyr4:
okay _now_ its ready for review 😅 
i intentionally moved the code around in `viewChunk` and added the function `processLogLine` so that there is room to add more per-chunk and per-line processing in future iterations. (for example, log timestamps and collapsible groups)

wass3rw3rk (approved):
lgtm

PR #674, opened 2023-06-09
fix(secrets): large names and general table render improvements 

plyr4:
pretty straightforward improvement to the way the `table-base` class renders secrets 

- adds a `.secret-name` class that sets a `max-width` of `15rem`. this helps fix the cutoff issue from large secrets
- renders "list cells" like `images` and `events` in a stack. this helps with the cutoff and makes images/events easier to read
- renders `key` and the "list cell elements" inside `<span>` elements to make them easy to copy/scroll. they are also given a max-width to help with the cutoff issues
- changes `no images` to `all images` to make it more clear what each secret is enabled for
- reuses the `-icon` class to make the table icon tighter

## before 
(notice cutoff from the large secret name)

Screenshot 2023-06-09 at 3 31 56 PM

## after

Screenshot 2023-06-09 at 3 30 39 PM


plyr4:
per feedback im going to look into converting the `<input>` to a styled scrollable `<span>`

plyr4:
converted the readonly `<input>` elements to `<span>` with `overflow: scroll`

ecrupper (approved):

timhuynh94 (approved):

PR #683, opened 2023-06-22
fix(schedules): correct encoding of update schedule payload for active field

ecrupper:
Calls to enable/disable an existing schedule in the UI do not work because the payload is encoded as `enabled: bool` rather than `active: bool`.


plyr4 (approved):

plyr4:
good catch. separate but related, should we change the wording in the UI to make it more accurate?
https://github.com/go-vela/ui/blob/main/src/elm/Pages/Schedules/View.elm#L130 etc

timhuynh94 (approved):

PR #689, opened 2023-07-06
enhance: add more fields to schedules table

plyr4:
Screenshot 2023-07-06 at 3 56 31 PM

also removing unused fields from the model that already exist in the form, like `id` `entry` and `enabled`


timhuynh94 (approved):
👌

PR #696, opened 2023-07-11
fix: apply overflow auto on list-item

plyr4:
should solve the overflow automatically showing on some browsers


timhuynh94 (approved):

KellyMerrick (approved):
lgtm

PR #707, opened 2023-07-18
feat(admin): add admin page containing a workers table

ecrupper:


JordanSussman:
@ecrupper I'm sure this is answered in the underlying server source code, but I'm curious: will this page only show workers that have checked in within the last x interval? I'm specifically asking since our workers table is quite "dirty" with old workers due to the amount of auto-scaling we do on a daily basis.

ecrupper:
> @ecrupper I'm sure this is answered in the underlying server source code, but I'm curious: will this page only show workers that have checked in within the last x interval? I'm specifically asking since our workers table is quite "dirty" with old workers due to the amount of auto-scaling we do on a daily basis.

Yeah something like that is definitely my vision for this (slated for `v0.23`). Maybe leveraging the `active` column? Check in logic works too. 

A few things I want to see most from this page: some sort of filtering on check-in logic / active status, ability to update routes, ability to click on running builds and be directed to that build's page

plyr4:
closing in favor of https://github.com/go-vela/ui/pull/794

PR #729, opened 2023-10-24
chore: rebuild package-lock

plyr4:
bumping various dependencies using
```
make clean
npm install
```

im just going to leave this here as a point of reference, the CI build is failing and a lot of things are bumped that i havent looked into

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,34 +108,97 @@
       ]
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.21.4",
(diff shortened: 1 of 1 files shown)


plyr4:
figured we would see this error
```
 Error: The expression evaluated to a falsy value:

  (0, _assert().default)(bundleGraph instanceof _BundleGraph.default)

  AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  
    (0, _assert().default)(bundleGraph instanceof _BundleGraph.default)
```

i've been seeing this locally

PR #730, opened 2023-10-26
enhance(build): logic parsing and linking of auto cancel messaging

ecrupper:
Types / Server PRs: https://github.com/go-vela/types/pull/299 & https://github.com/go-vela/server/pull/911

UI essentially determines whether a build was canceled or auto-canceled based on the content of the error message and then links the superseding build in the message itself.

Will update description with styling decisions when made and bring the PR out of draft at that point.


docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
   # https://go-vela.github.io/docs/administration/server/
   server:
     container_name: server
-    image: target/vela-server:latest
+    image: server:local
```
wass3r, line 37: suggestion: not sure if this was intended to be checked in, but i'd personally prefer to revert this. it assumes a certain work style. for example, if i want to pair local dev with certain server code, i temporarily point this at the local server repo to live build the image. leaving it as it was will typically allow folks to get the local stack running without having to follow extra steps.

wass3r (approved):
Feels like cheating but I happen to approve :) nice work.

claire1618 (approved):
seems good to me!

plyr4 (approved):
<3

PR #740, opened 2023-11-06
fix(viz): track draw state to apply recenter on fresh draws

plyr4:
more stability for the elm vs. js render race condition, where the page root doesnt render quick enough for the graph render dispatch

also 
- renames `isRefreshDraw` to `sameBuild` to make it a bit clearer
- renames `centerOnDraw` to `freshDraw` to make it a bit clearer

### DIFF
--- a/cypress/fixtures/build_graph.json
+++ b/cypress/fixtures/build_graph.json
@@ -1,5 +1,8 @@
 {
   "build_id": 4,
+  "build_number": 4,
+  "org": "github",
+  "repo": "octocat",
   "nodes": {
     "0": {
       "id": 0,
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -1071,7 +1071,7 @@ update msg model =
                                 um_ =
                                     updateRepoModels model rm bm ugm
                             in
-                            ( ugm, Cmd.batch [ renderBuildGraph um_ False ] )
+                            ( ugm, renderBuildGraph um_ False )
 
                         _ ->
                             ( model.repo.build.graph, Cmd.none )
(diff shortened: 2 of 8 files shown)


wass3rw3rk (dismissed):

wass3r (dismissed):
~seems to work now. i'm not able to reproduce the issue.~
was working on another issue and have to backtrack here. i can reliably reproduce with following steps:
1. visit repo builds page
1. click on any build to go to its build page
1. click "Visualize" tab

it works fine when i hard refresh repo builds page after step 1. and do the remaining steps

wass3r:
fwiw, i'm able to get pretty consistent results it seems by changing two things:
1. use `renderBuildGraph updatedModel True` here https://github.com/go-vela/ui/blob/3c4fe3c5d3d32c5853b97291fd6657234a16ad5a/src/elm/Main.elm#L2169
2. change the following to use `renderBuildGraph um True` https://github.com/go-vela/ui/blob/3c4fe3c5d3d32c5853b97291fd6657234a16ad5a/src/elm/Main.elm#L4370 

in other words, passing a truthy `centerOnDraw` more.

worst result i saw is an initial flash of uncentered graph and then it fixed itself. seemed to be pretty rare - more rare than the initial issues, i think?!

on other minor optimization with this would be to keep track of whether it has already centered and zoomed because it might call it multiple times depending on mix of isrefreshdraw and centerondraw.

didn't notice any ill side effects otherwise, but it's possible there's some use case i'm not accounting for. was just trying a few things :)

plyr4:
@wass3r thank you for taking a look 🙏 i appreciate it

> use `renderBuildGraph updatedModel True` here 
im glad we're seeing the same thing. that change should be present in this PR and it fixed most of the issues i was seeing.

i'll make the second change, good call

wass3rw3rk (dismissed):

wass3rw3rk (approved):

timhuynh94 (approved):

PR #751, opened 2023-12-29
feat(repo_settings): new allow_events support

ecrupper:
Final piece of the puzzle which was started with [this proposal](https://github.com/go-vela/community/pull/828)

A little tricky to encode / decode nested JSON while also taking an all-or-nothing approach with the AllowEvents struct. All in all, very little functional differences here, but there's a much more visibility to the users as to the type of webhooks that Vela is receiving. 

Here's a screenshot of the new look:

Screenshot 2023-12-29 at 3 34 45 PM


src/elm/Main.elm
```diff
@@ -433,7 +434,7 @@ type Msg
     | DisableRepo Repository
     | ChownRepo Repository
     | RepairRepo Repository
-    | UpdateRepoEvent Org Repo Field Bool
+    | UpdateRepoEvent Org Repo Repository Field Bool
```
wass3r, line 437: that can get confusing :D might have to come up with better naming, but can be a separate PR

wass3r (approved):

claire1618 (approved):

PR #753, opened 2024-01-04
feat(deployments): updating ui to show new changes to database

claire1618:
This story is attached to this PR: [https://github.com/go-vela/server/pull/1031](https://github.com/go-vela/server/pull/1031)

In switching over from pulling in deployments information from the SCM to pulling in from our own database, a few changes were made in the ui to allow this to happen.

As well, an ID column and builds column were added to the deployments table to better help users comprehend and use deployments.

### DIFF
--- a/src/elm/Pages/Deployments/View.elm
+++ b/src/elm/Pages/Deployments/View.elm
@@ -275,6 +275,7 @@ redeployLink org repo deployment =
         [ text "Redeploy"
         ]
 
+
 {-| pullBuildLinks : takes deployment and creates a list of links to every build in the builds field
 -}
 pullBuildLinks : Deployment -> List String
@@ -293,7 +294,7 @@ linksView : List String -> Html msg
 linksView links =
     links
         |> List.map (\link -> a [ href link ] [ text (String.append "#" (getHead (List.head (List.reverse (String.split "/" link))))) ])
-        |> List.intersperse ( text ", " )
+        |> List.intersperse (text ", ")
         |> div []
 
 
@@ -304,6 +305,6 @@ getHead link =
     case link of
         Nothing ->
             ""
-        
+
         Just val ->
             val
--- a/src/elm/Pages/Deployments/View.elm
+++ b/src/elm/Pages/Deployments/View.elm
@@ -275,6 +275,7 @@ redeployLink org repo deployment =
         [ text "Redeploy"
         ]
 
+
 {-| pullBuildLinks : takes deployment and creates a list of links to every build in the builds field
 -}
 pullBuildLinks : Deployment -> List String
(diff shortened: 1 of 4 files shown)


KellyMerrick (approved):
lgtm

src/elm/Pages/Deployments/View.elm
```diff
@@ -258,3 +274,37 @@ redeployLink org repo deployment =
...
+{-| pullBuildLinks : takes deployment and creates a list of links to every build in the builds field
+-}
+pullBuildLinks : Deployment -> List String
+pullBuildLinks deployment =
+    case deployment.builds of
+        Nothing ->
+            []
+
+        Just builds ->
+            List.map (String.dropRight 1) (List.map (String.dropLeft 1) (List.map Debug.toString (List.map .link builds)))
```
wass3r, line 288: `Debug.` shouldn't be included in code. Is that leftover from debugging? Should this just be `List.map .link builds`?

src/elm/Pages/Deployments/View.elm
```diff
@@ -258,3 +274,37 @@ redeployLink org repo deployment =
...
+
+
+{-| linksView : takes list of links and creates an HTML msg that displays as a list of links
+-}
+linksView : List String -> Html msg
+linksView links =
+    links
+        |> List.map (\link -> a [ href link ] [ text (String.append "#" (getHead (List.head (List.reverse (String.split "/" link))))) ])
```
wass3r, line 296: Suggestion: rewrite the build number + "#" construction to use `|>` (pizza slices). i find it more readable.
```elm
link
  |> String.split "/"
  |> List.reverse
  |> List.head
  |> Maybe.withDefault ""
  |> String.append "#"
```
(using `Maybe.withDefault` we can also drop the `getHead` function)

Another thought is to leverage [Route.match](https://github.com/go-vela/ui/blob/c1c1e1701c18b8063162f2a956f44eb5624fb764/src/elm/Routes.elm#L102) to map the string to a build link and then extract the right part. Then we get some validation that the provided link matches a build link pattern.

wass3r (changes requested):

src/elm/Pages/Deployments/View.elm
claire1618: yes it should be! thank you!

src/elm/Pages/Deployments/View.elm
claire1618: I went with Maybe.withDefault and was able to clean up the code quite a bit! I can look into Route.match if you think that would be better

wass3r (approved):
i noticed we don't have tests for the view even prior to this PR? we should add a story to follow up after this to add them.

timhuynh94 (approved):

PR #756, opened 2024-01-17
enhance(secrets): update UI decode / encode / payload building for AllowEvents

ecrupper:
I was able to reuse _some_ of the code I wrote for repo allow events, but not all, which was unfortunate. Wish I could make this more efficient.

Here's a screen capture of what the new UI will look like:

Screenshot 2024-01-17 at 11 44 12 AM

Screenshot 2024-01-17 at 11 45 44 AM

### DIFF
--- a/src/scss/_forms.scss
+++ b/src/scss/_forms.scss
@@ -261,6 +261,16 @@
     }
   }
 
+  &.-two-col-secrets {
+    flex-flow: row wrap;
+    gap: 0.5rem;
+    max-width: 30%;
+
+    .form-control {
+      flex: 0 0 40%;
+    }
+  }
+
   .form-control {
     margin-right: 0.5rem;
   }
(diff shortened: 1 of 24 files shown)


src/scss/_forms.scss
```diff
@@ -261,6 +261,16 @@
     }
   }
 
+  &.-two-col-secrets {
+    flex-flow: row wrap;
+    gap: 0.5rem;
+    max-width: 30%;
```
wass3rw3rk, line 267: i'll check on this, but not a fan to create a one-off class here. that means the original is not flexible enough to handle this scenario.

claire1618 (approved):

src/scss/_forms.scss
wass3rw3rk: i'll open separate PR

wass3rw3rk (approved):

PR #757, opened 2024-01-24
fix: removing id field and adding created at field to deployments table

claire1618:
Removing the ID field because it doesn't really provide any useful data.
Adding the Created At field so that users can see when deployments were created.

Going from:
Screenshot 2024-01-24 at 10 19 19 AM

to: 
Screenshot 2024-01-24 at 10 18 09 AM

### DIFF
--- a/src/elm/Main.elm
+++ b/src/elm/Main.elm
@@ -64,7 +64,7 @@ import Html.Attributes
         , type_
         )
 import Html.Events exposing (onClick)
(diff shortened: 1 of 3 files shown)


ecrupper (approved):
Nice

KellyMerrick (approved):
lgtm

plyr4 (approved):

timhuynh94 (approved):

PR #759, opened 2024-01-29
feat: adding a delete event

claire1618:
Similar to the pull_request and push events, it would be beneficial if Vela could recognize the delete event. So the users would have the option to have Vela set off builds when a branch or tag are deleted in their repo.

The code here adds the Delete event to the checkboxes in Secrets and RepoSettings.

Goes along with: [https://github.com/go-vela/server/pull/1038](https://github.com/go-vela/server/pull/1038)


timhuynh94 (approved):

ecrupper (approved):

src/elm/Vela.elm
```diff
@@ -2503,7 +2517,7 @@ secretToKey secret =
 
 defaultSecretAllowEvents : AllowEvents
 defaultSecretAllowEvents =
-    { push = { branch = True, tag = True }, pull = defaultPullActionsPayload Nothing, deploy = { created = True }, comment = defaultCommentActionsPayload Nothing, schedule = defaultScheduleActionsPayload Nothing }
+    { push = { branch = True, tag = True, delete_branch = True, delete_tag = True }, pull = defaultPullActionsPayload Nothing, deploy = { created = True }, comment = defaultCommentActionsPayload Nothing, schedule = defaultScheduleActionsPayload Nothing }
```
ecrupper, line 2520: I'd say delete events should not be part of the default set for secrets. Let's just keep it to push, tag, and deployment 👍

ecrupper (changes requested):

ecrupper (approved):

wass3rw3rk (approved):

PR #772, opened 2024-02-29
fix(deployments): remove unnecessary class

wass3r:
fixes a styling issue in light mode on the deployments page (see erroneous white lines).

before
before

after
after

also tested when the column was populated with builds, fwiw


plyr4 (approved):

KellyMerrick (approved):
lgtm

timhuynh94 (approved):

PR #779, opened 2024-03-26
enhance: elm-land opinionated tweaks

plyr4:
trying to move towards being able to use the `elm-land` CLI to generate boiler plate code.

progress towards https://github.com/go-vela/community/issues/961

- chore: use appropriate elm-land module names that match routes
- chore: move helper out of generated file

these changes let `elm-land generate` run without error, though the CLI still doesnt generate files in a way that we want, so its a work in progress.  

note on the 122 removed lines: the PR also deletes `PageTemplate.elm` since you can easily do `elm-land add page` now and it spits out the exact same file, more or less


KellyMerrick (dismissed):
lgtm

KellyMerrick (approved):
lgtm

ecrupper (approved):

PR #795, opened 2024-04-26
feat(schedule)!: show schedule errors and nest object

wsan3:
### Description
Whenever a user has a YAML bug in their pipeline, or any other build-breaking error that short-circuits pipeline execution, they can usually find these error in the Audit tab, where we populate the hook with the error message.

This is currently not available for schedules. Instead, the error is just logged.

### Desired View
Screenshot 2024-04-26 at 10 37 57 AM

### Related
https://github.com/go-vela/community/issues/872

### DIFF
--- a/src/elm/Pages/Org_/Repo_/Schedules.elm
+++ b/src/elm/Pages/Org_/Repo_/Schedules.elm
@@ -16,8 +16,8 @@ import Components.Table
 import Dict
 import Effect exposing (Effect)
 import FeatherIcons
-import Html exposing (Html, a, code, div, span, text, td, tr)
-import Html.Attributes exposing (class, attribute)
+import Html exposing (Html, a, code, div, span, td, text, tr)
+import Html.Attributes exposing (attribute, class)
 import Http
 import Http.Detailed
 import Layouts
@@ -296,6 +296,7 @@ schedulesToRows zone org repo schedules =
         |> List.concatMap (\s -> [ Just <| Components.Table.Row (addKey s) (viewSchedule zone org repo), scheduleErrorRow s ])
         |> List.filterMap identity
 
+
 {-| tableHeaders : returns table headers for schedules table
 -}
 tableHeaders : Components.Table.Columns
@@ -403,14 +404,15 @@ viewScheduleError schedule =
                 |> List.filterMap identity
 
         msgRow =
-                tr [ class "error-data", Util.testAttribute "schedules-error" ]
-                    [ td [ attribute "colspan" "6" ]
-                        [ code [ class "error-content" ]
-                            lines
-                        ]
+            tr [ class "error-data", Util.testAttribute "schedules-error" ]
+                [ td [ attribute "colspan" "6" ]
+                    [ code [ class "error-content" ]
+                        lines
                     ]
+                ]
     in
     msgRow
+
 
 {-| addKey : helper to create Vela.Schedule key
 -}
--- a/src/elm/Pages/Org_/Repo_/Schedules.elm
+++ b/src/elm/Pages/Org_/Repo_/Schedules.elm
@@ -293,7 +293,7 @@ viewRepoSchedules shared model org repo =
 schedulesToRows : Time.Zone -> String -> String -> List Vela.Schedule -> Components.Table.Rows Vela.Schedule msg
 schedulesToRows zone org repo schedules =
     schedules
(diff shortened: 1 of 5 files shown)


plyr4:
whats the plan for adding the other changes made in https://github.com/go-vela/server/pull/1108 like [nested repos](https://github.com/go-vela/server/pull/1108/files#diff-82b93a6580e99aa91100875d8ffe9907b3ac95f52be235843e383b139b824277R14)? 
will those be in their own PR or attached to this one?

wsan3:
> whats the plan for adding the other changes made in [go-vela/server#1108](https://github.com/go-vela/server/pull/1108) like [nested repos](https://github.com/go-vela/server/pull/1108/files#diff-82b93a6580e99aa91100875d8ffe9907b3ac95f52be235843e383b139b824277R14)? will those be in their own PR or attached to this one?

I'll attach the nested Schedule and types migration changes to this PR

src/elm/Pages/Org_/Repo_/Schedules.elm
```diff
@@ -376,8 +381,34 @@ viewSchedule zone org repo schedule =
...
+                    (\line ->
+                        Just <|
+                            Ansi.Log.viewLine line
+                    )
+                |> Array.toList
+                |> List.filterMap identity
+
+        msgRow =
+            tr [ class "error-data", Util.testAttribute "schedules-error" ]
+                [ td [ attribute "colspan" "6" ]
```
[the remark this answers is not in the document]
plyr4: im just nitpicking. i realize now that its hardcoded in some places but i was thinking about this spot
https://github.com/go-vela/ui/blob/main/src/elm/Components/Table.elm#L131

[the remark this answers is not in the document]
wsan3: I took that snippet from `viewHookError` in `src/elm/Pages/Org_/Repo_/Hooks.elm`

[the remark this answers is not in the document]
plyr4: sweet, figured 👍

plyr4 (approved):

src/elm/Pages/Org_/Repo_/Schedules.elm
```diff
@@ -376,8 +381,34 @@ viewSchedule zone org repo schedule =
...
+    let
+        lines =
+            Utils.Ansi.decodeAnsi schedule.error
+                |> Array.map
+                    (\line ->
+                        Just <|
+                            Ansi.Log.viewLine line
+                    )
+                |> Array.toList
+                |> List.filterMap identity
```
wass3r, line 404: do we actually need to deal with Ansi here? it would always be straight text, correct?

wsan3: Good point. Yeah, the errors should show as plain text. I copied what was used for `viewHookError` and applied it to `viewScheduleError`. Should I remove `decodeAnsi ` from both? @plyr4 What are your thoughts?

plyr4: hook errors can contain ansi, see: https://github.com/go-vela/ui/pull/305
though, maybe that was added in the small timeframe where we switched over to `goccy/yaml` that was producing pretty errors. i cant remember, it was 3 years ago 😅

plyr4: lol yup the timing is lining up, here's the PR where we reverted back to the non-pretty compiler errors:

https://github.com/go-vela/compiler/pull/133

we just never reverted the ui changes that decode the pretty errors

plyr4: so its more likely that its not needed here, nor the audit/hooks page 😅 
but if we ever try to improve the compiler error messages..... then it will be awesome to have! 
so its up to yall on whether you want to remove it here, remove it in both, leave it, whatever

wass3r: makes sense, hopefully we get there to the pretty errors soon 😅 i'm ok either way, i don't think the overhead is bad, especially for this., since it's already removed, we will go with that, i suppose. thanks @wsan3 @plyr4

wass3r (dismissed):
thanks

ecrupper (approved):

plyr4 (approved):

PR #810, opened 2024-08-02
fix: reduce excessive API calls

wass3r:
reduces api calls on build, services, graph, org, secrets, and repo pages 
- reduces unnecessary calls once a build is finished
- eliminates duplicate or redundant calls
- reverts interval on graph api requests back to 5s as was the case in 0.24. 1s graph rerenders will still be in place
- handle 401s due to token expiration on non /token-refresh calls better


KellyMerrick (approved):
lgtm

plyr4 (approved):

KellyMerrick (approved):
lgtm

PR #813, opened 2024-08-07
fix: stop double log fetch on steps/services

plyr4:
while testing #810 i noticed that logs are called twice when you click on a step/service header, this is unintended side effect of tracking `OnHashChanged` events and using that to fetch logs without an extra check

does not impact refreshes, but still better than doing double calls for no reason.


KellyMerrick (dismissed):
lgtm

plyr4:
putting in draft temporarily, adding some more improvements

wass3r (approved):

